PR #25, opened 2017-02-07
Issue #22: Add a --quiet flag for the b cron command.

### LINKED ISSUE #22: Add a --quiet option for bee commands.
What I had in mind was something like the `if($type == 'log' && !b_is_debug())` filter in `includes/common.inc` (around line 84) that would similarly filter out status, success and log messages.

The goal is to have the `--quiet`option either suppress all output (like drush) and note success or failure through the return value in $?, or they reduce the output, removing the progress and status messages, but leaving any warnings or errors. Some commands (mplayer, for example) support multiple levels of quiet, like multiple levels of -v for verbosity.

For example using `b` to trigger cron is a common use case.  The current version of `b` will print out a success message if everything works, which when run through cron creates a lot of emails.

We should add a `--quiet` (`-q`) option that doesn't print out any of the success or log messages, just the errors and warnings.  Perhaps two `-q` options could suppress the warnings, too.

serundeputy:
This fixes #22.

### DIFF
--- a/commands/Core/cron.b.inc
+++ b/commands/Core/cron.b.inc
@@ -9,6 +9,9 @@
   $items['cron'] = array(
     'bootstrap' => B_BOOTSTRAP_FULL,
     'description' => 'Process cron.',
+    'options' => array(
+      'quiet' => 'Supress output messages for non-interactive modes.',
+    ),
     'callback' => 'cron_command_process',
   );
   return $items;
@@ -17,5 +20,7 @@
 
 function cron_command_process($arguments, $options) {
   backdrop_cron_run();
-  b_set_message('Cron processed.', 'success');
+  if (!isset($options['quiet'])) {
+    b_set_message('Cron processed.', 'success');
+  }
 }


jlfranklin (changes requested):
The patch does address the issue as described by the title ("add --quiet for the cron command"), but the body of the issue is looking for something a bit more general, and used the cron command as an example of when the `--quiet` flag would be useful.

What I had in mind was something like the `if($type == 'log' && !b_is_debug())` filter in `includes/common.inc` (around line 84) that would similarly filter out `status`, `success` and `log` messages.

Limiting the `--quiet` to _only_ the suppress the `cron` command is not the way most command line utilities treat a quiet flag.  Usually, they either suppress all output (like `drush`) and note success or failure through the return value in `$?`, or they reduce the output, removing the progress and status messages, but leaving any warnings or errors.  Some commands (`mplayer`, for example) support multiple levels of quiet, like multiple levels of `-v` for verbosity.

Can the patch be more general?

serundeputy:
I see.  Yes, that is a better idea.
Let's go for the general approach.

thanks,
~Geoff

PR #35, opened 2019-10-03
Issue #12: add multisite support.

### LINKED ISSUE #12: b doesn't support multisite
B has no way to say `b --root /var/www/backdrop --url www.example.org` to select `sites/example.org/settings.php` instead of `/var/www/backdrop/settings.php`.

---

After some time of silence here, we now have several options:

@jlfranklin's PR: https://github.com/backdrop-contrib/b/pull/21
~@indigoxela's PR: https://github.com/backdrop-contrib/b/pull/35~ Deprecated in favour of https://github.com/backdrop-contrib/b/pull/37
@BWPanda's PR: https://github.com/backdrop-contrib/b/pull/37

indigoxela:
Consider the sites.php file to figure out the backdrop root. Some smaller whitespace fixes.

Fixes https://github.com/backdrop-contrib/b/issues/12

### DIFF
--- a/b.php
+++ b/b.php
@@ -87,6 +87,8 @@ function b_init() {
       }
     }
   }
+
+  b_init_blobals($hostname);
   
   if ($_backdrop_root) {
     define('BACKDROP_ROOT', $_backdrop_root);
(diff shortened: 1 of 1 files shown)


indigoxela:
Ooops.. This needs some more work. It turned out, it definitely needs more than just HTTP_HOST to make b happy. (done)

ghost:
Good job, and kudos on such a simple set of changes!

Unfortunately I noticed the following issues:
- No support for downloading to a multisite directory (downloads still go to `[root]/modules`)
- `b` can only be run from directories with a `settings.php` file (e.g. not `[root]/sites/[site]/modules`)

So your PR is good at what it does, just a bit limited in terms of fully supporting multisites. Since you suggested in Gitter that my PR was a better approach, do you want to work together on that and deprecate this PR in its favour?

indigoxela:
> do you want to work together on that and deprecate this PR in its favour?

Absolutely. Let's combine forces. ;)

PR #37, opened 2019-10-03
Issue #12: Provide full support for multisites.

### LINKED ISSUE #12: b doesn't support multisite

ghost:
Fixes https://github.com/backdrop-contrib/b/issues/12

### DIFF
--- a/b.php
+++ b/b.php
@@ -51,6 +51,7 @@ function b_init() {
   $command = array(
    'options' => array(
      'root' => 'Backdrop root folder',
+     'url' => 'Backdrop site URL (as defined in sites.php)',
      'drush' => 'Use .drush.inc files instead. Drupal 7 drush commands compatibility.',
      'y' => 'Force Yes to all Yes/No questions',
      'yes' => 'Force Yes to all Yes/No questions',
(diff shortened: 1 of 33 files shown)


indigoxela:
Nice work! But I do have one concern.

Currently `b dl` forces downloading to the main directory (modules, themes...) You now enforce downloading to the sites subdirs if it's a multisite install.

Both doesn't seem perfect to me.

If I use multisite, I always share at least the majority of contrib modules (hence main dir is fine). Only single modules and likely the theme or some layouts are site specific.
So your approach actually makes it more complicated for me to use "b" with multisite.

ghost:
@indigoxela Thanks for the feedback! It's good to hear how other people use Backdrop/multisite.

So with this PR if you run `b dl` from within your site directory, it'll download to there (e.g. `/sites/[SITE]/modules`). Whereas if you run the same command from outside of a site directory (e.g. `/backdrop`), it'll download to there instead (e.g. `/backdrop/modules`). So basically it should download to wherever you run this from.

I believe this is how Drush for Drupal 7 works. Drush for Backdrop doesn't fully support multisites yet, which is why I started using `b` instead as it's much easier to fix :smiley: 

How would you prefer it to work? E.g. where do you prefer to run commands from, and where do you want them to download to?

ghost:
BTW @indigoxela, are you on Gitter? https://gitter.im/backdrop/backdrop-issues If so, we can chat there if it's easier.

indigoxela:
> it should download to wherever you run this from.

Ah, I didn't test that from the backdrop root, as "b" says "BackdropCMS is not installed yet." - but downloads it then to the main dirs. Cool, very helpful!

The warning message was misleading me.

BTW: Drush with D7 allows downloading projects to anywhere, regardless of an install.

ghost:
> The warning message was misleading me.

Yeah, that message is displayed if `b` can't find a `settings.php` file and therefore there's technically no Backdrop site available. But in that case it can (and should IMO) still download things to the root directory. Maybe a separate issue can address the wording/display of that message...

b.php
```diff
@@ -61,34 +61,43 @@ function b_init() {
...
+
     chdir($_backdrop_root);
     $full_path = getcwd();
     define('BACKDROP_ROOT', $full_path);
+    $_SERVER['HTTP_HOST'] = basename($full_path);
+    if ($_backdrop_site) {
+      define('BACKDROP_SITE', $_backdrop_site);
+      $_SERVER['HTTP_HOST'] = $_backdrop_site;
+    }
     require_once 'core/includes/bootstrap.inc';
```
indigoxela: Wouldn't this fail, if one calls b from a multisite directory (which includes the sites settings.php).

I think, we should make this (all of) b's _require_once_ more robust regarding the base dir for including core files.

indigoxela: If in one of the multisite directories containing its' settings.php `$_backdrop_root` gets set to the current directory, not the main Backdrop directory. Hence including bootstrap.inc from core fails (wrong path).

If in the main directory, the path is correct, but "BackdropCMS is not installed yet.".

Something from your previous solution seems to be missing, as I know this has been working correctly then.

ghost: Thanks for the review, I'll take a look tonight.

ghost: @indigoxela Can you tell me exactly what command you're running and from which directory?

If I run `b st` from `backdrop/sites/my-site`, I get:

`Backdrop CMS Installation detected` (i.e. it found `settings.php`)
`Backdrop root: backdrop`
`Backdrop site: my-site`

PR #44, opened 2019-10-10
Issue #43: Display appropriate message based on cron success.

### LINKED ISSUE #43: Don't display success message if cron doesn't run
Currently, a success message is displayed regardless of whether cron actually runs when using the `b cron` command. We should check the return status of `backdrop_cron_run()` and display a message appropriately.

ghost:
Fixes https://github.com/backdrop-contrib/b/issues/43

### DIFF
--- a/commands/Core/cron.b.inc
+++ b/commands/Core/cron.b.inc
@@ -16,6 +16,11 @@
 
 
 function cron_command_process($arguments, $options) {
-  backdrop_cron_run();
-  b_set_message('Cron processed.', 'success');
+  $success = backdrop_cron_run();
+  if ($success) {
+    b_set_message('Cron processed.', 'success');
+  }
+  else {
+    b_set_message('Cron is already running.', 'error');
+  }
 }


commands/Core/cron.b.inc
```diff
@@ -16,6 +16,11 @@
...
 
 function cron_command_process($arguments, $options) {
-  backdrop_cron_run();
-  b_set_message('Cron processed.', 'success');
+  $success = backdrop_cron_run();
+  if ($success) {
+    b_set_message('Cron processed.', 'success');
+  }
+  else {
+    b_set_message('Cron is already running.', 'error');
```
indigoxela, line 24: That's probably not the only cause for backdrop_cron_run() returning false.
Think of a severe problem on the site, running into a timeout, some eventually broken cron hook in a module...

ghost: Hmm, I got that from the return value of [`backdrop_cron_run()`](https://api.backdropcms.org/api/backdrop/core%21includes%21common.inc/function/backdrop_cron_run/1) which says:

> TRUE if cron ran successfully and FALSE if cron is already running.

If that's not necessarily the case, then we should open a issue against core to change that wording to be more accurate...

indigoxela: Interesting point... Actually backdrop_cron_run() uses try/catch for module's cron hooks.

While I'm not 100% sure, if the docs are correct regarding failure, assuming so is fine for now (I guess).

ghost: Yeah, but doesn't look like that affects the return value... If it gets to that stage of calling other hook implementations, it looks like it just returns TRUE regardless.

PR #135, opened 2021-03-03
Add php-eval command.

### LINKED ISSUE #128: Add `php-eval` command
`php-eval` "my_function()" 

from drush would be really helpful. It lets you accomplish almost anything, sort of the tool of last resort.

Aliases ev, eval, php:eval

bugfolder:
Add the php-eval command (aliases ev, eval).

In response to [this issue](https://github.com/backdrop-contrib/b/issues/128).

### DIFF
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -35,7 +35,10 @@ function php_eval_b_callback($arguments, $options) {
   }
   catch(ParseError $e) {
     // This is more readable than the default error we would get from PHP.
-    $err_msg = $e->getMessage() . ' ' . bt('in:') . ' ' . $arguments['code'];
+    $err_msg = bt('@error in: @code', array(
+      '@error' => $e->getMessage(),
+      '@code' => $arguments['code'],
+    ));
     b_message($err_msg, 'error');
   }
 }
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -31,7 +31,7 @@ function eval_b_command() {
  */
 function php_eval_b_command($arguments, $options) {
   try {
-    eval($arguments['code']);
+    eval($arguments['code'] . ';');
   }
   catch(ParseError $e) {
     // This is more readable than the default error we would get from PHP.
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -29,7 +29,7 @@ function eval_b_command() {
 /**
  * Command callback: Execute specified PHP code.
  */
-function php_eval_b_command($arguments, $options) {
+function php_eval_b_callback($arguments, $options) {
   try {
     eval($arguments['code'] . ';');
   }
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -18,16 +18,16 @@ function eval_b_command() {
       'aliases' => array('ev', 'eval'),
       'bootstrap' => B_BOOTSTRAP_FULL,
       'examples' => array(
-        'b php-eval \'$node = node_load(1); print $node->title;\'' => 'Loads node with nid 1 and then prints its title.',
-        'b php-eval "node_access_rebuild();"' => 'Rebuild node access permissions.',
-        'b php-eval "file_unmanaged_copy(\'$HOME/Pictures/image.jpg\', \'public://image.jpg\');"' => 'Copies a file whose path is determined by an environment\'s variable. Note the use of double quotes so the variable $HOME gets replaced by its value.',
+        'b php-eval \'$node = node_load(1); print $node->title;\'' => bt('Loads node with nid 1 and then prints its title.'),
+        'b php-eval "node_access_rebuild();"' => bt('Rebuild node access permissions.'),
+        'b php-eval "file_unmanaged_copy(\'$HOME/Pictures/image.jpg\', \'public://image.jpg\');"' => bt('Copies a file whose path is determined by an environment\'s variable. Note the use of double quotes so the variable $HOME gets replaced by its value.'),
       ),
     ),
   );
 }
 
 /**
- * Command callback: Execute specified PHP code.
+ * Command callback: Evaluate arbitrary PHP code after bootstrapping Backdrop.
  */
 function php_eval_b_callback($arguments, $options) {
   try {
@@ -35,9 +35,9 @@ function php_eval_b_callback($arguments, $options) {
   }
   catch(ParseError $e) {
     // This is more readable than the default error we would get from PHP.
-    $err_msg = bt('@error in: @code', array(
-      '@error' => $e->getMessage(),
-      '@code' => $arguments['code'],
+    $err_msg = bt('!msg in: !code', array(
+      '!msg' => $e->getMessage(),
+      '!code' => $arguments['code'],
     ));
     b_message($err_msg, 'error');
   }
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -27,7 +27,8 @@ function eval_b_command() {
 }
 
 /**
- * Command callback: Evaluate (run/execute) arbitrary PHP code after bootstrapping Backdrop.
+ * Command callback: Evaluate (run/execute) arbitrary PHP code after
+ * bootstrapping Backdrop.
  */
 function php_eval_b_callback($arguments, $options) {
   try {
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -10,7 +10,7 @@
 function eval_b_command() {
   return array(
     'php-eval' => array(
-      'description' => bt('Evaluate arbitrary PHP code after bootstrapping Backdrop.'),
+      'description' => bt('Evaluate (run/execute) arbitrary PHP code after bootstrapping Backdrop.'),
       'callback' => 'php_eval_b_callback',
       'arguments' => array(
         'code' => bt('The PHP code to evaluate.'),
@@ -27,7 +27,7 @@ function eval_b_command() {
 }
 
 /**
- * Command callback: Evaluate arbitrary PHP code after bootstrapping Backdrop.
+ * Command callback: Evaluate (run/execute) arbitrary PHP code after bootstrapping Backdrop.
  */
 function php_eval_b_callback($arguments, $options) {
   try {
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -1,0 +1,41 @@
+<?php
+/**
+ * @file
+ * Command(s) for evaluating PHP code directly.
+ */
+
+/**
+ * Implements hook_b_command().
+ */
+function eval_b_command() {
+  return array(
+    'php-eval' => array(
+      'description' => bt('Evaluate arbitrary PHP code after bootstrapping Backdrop.'),
+      'callback' => 'php_eval_b_command',
+      'arguments' => array(
+        'code' => bt('The PHP code to evaluate.'),
+      ),
+      'aliases' => array('ev', 'eval'),
+      'bootstrap' => B_BOOTSTRAP_FULL,
+      'examples' => array(
+        'b php-eval \'$node = node_load(1); print $node->title;\'' => 'Loads node with nid 1 and then prints its title.',
+        'b php-eval "node_access_rebuild();"' => 'Rebuild node access permissions.',
+        'b php-eval "file_unmanaged_copy(\'$HOME/Pictures/image.jpg\', \'public://image.jpg\');"' => 'Copies a file whose path is determined by an environment\'s variable. Note the use of double quotes so the variable $HOME gets replaced by its value.',
+      ),
+    ),
+  );
+}
+
+/**
+ * Command callback: Execute specified PHP code.
+ */
+function php_eval_b_command($arguments, $options) {
+  try {
+    eval($arguments['code']);
+  }
+  catch(ParseError $e) {
+    // This is more readable than the default error we would get from PHP.
+    $err_msg = $e->getMessage() . ' ' . bt('in:') . ' ' . $arguments['code'];
+    b_message($err_msg, 'error');
+  }
+}
--- a/commands/eval.b.inc
+++ b/commands/eval.b.inc
@@ -11,7 +11,7 @@ function eval_b_command() {
   return array(
     'php-eval' => array(
       'description' => bt('Evaluate arbitrary PHP code after bootstrapping Backdrop.'),
-      'callback' => 'php_eval_b_command',
+      'callback' => 'php_eval_b_callback',
       'arguments' => array(
         'code' => bt('The PHP code to evaluate.'),
       ),


commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+ * Command(s) for evaluating PHP code directly.
+ */
+
+/**
+ * Implements hook_b_command().
+ */
+function eval_b_command() {
+  return array(
+    'php-eval' => array(
+      'description' => bt('Evaluate arbitrary PHP code after bootstrapping Backdrop.'),
```
ghost: Remove reference to bootstrapping Backdrop (see note below about bootstrap). Also explain 'evaluate', since it sounds like this command is meant to *review* PHP code...

```suggestion
      'description' => bt('Evaluate (run/execute) arbitrary PHP code.'),
```

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+      'callback' => 'php_eval_b_command',
```
ghost: Change 'command' to 'callback' for consistency.

```suggestion
      'callback' => 'php_eval_b_callback',
```

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+        'b php-eval "file_unmanaged_copy(\'$HOME/Pictures/image.jpg\', \'public://image.jpg\');"' => 'Copies a file whose path is determined by an environment\'s variable. Note the use of double quotes so the variable $HOME gets replaced by its value.',
+      ),
+    ),
+  );
+}
+
+/**
+ * Command callback: Execute specified PHP code.
+ */
+function php_eval_b_command($arguments, $options) {
```
ghost: Change 'command' to 'callback' for consistency.

```suggestion
function php_eval_b_callback($arguments, $options) {
```

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+function eval_b_command() {
+  return array(
+    'php-eval' => array(
+      'description' => bt('Evaluate arbitrary PHP code after bootstrapping Backdrop.'),
+      'callback' => 'php_eval_b_command',
+      'arguments' => array(
+        'code' => bt('The PHP code to evaluate.'),
+      ),
+      'aliases' => array('ev', 'eval'),
+      'bootstrap' => B_BOOTSTRAP_FULL,
```
ghost, line 19: Is this really necessary? What if I wanted to check the PHP version installed on my system without having a Backdrop site setup yet? Including this means generic commands can't be run without a full Backdrop site, whereas removing it means they can, but doesn't prevent commands interacting with Backdrop when available (i.e. the best of both worlds). So I'd suggest removing this.

```suggestion
```

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+      'examples' => array(
+        'b php-eval \'$node = node_load(1); print $node->title;\'' => 'Loads node with nid 1 and then prints its title.',
```
ghost: Is there a reason to surround the PHP code with single quotes and then escape them? Seems it'd just be easier to use double-quotes instead... Also, example descriptions should be translatable.

```suggestion
        'b php-eval "$node = node_load(1); print $node->title;"' => bt('Loads node with nid 1 and then prints its title.'),
```

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+        'b php-eval "node_access_rebuild();"' => 'Rebuild node access permissions.',
+        'b php-eval "file_unmanaged_copy(\'$HOME/Pictures/image.jpg\', \'public://image.jpg\');"' => 'Copies a file whose path is determined by an environment\'s variable. Note the use of double quotes so the variable $HOME gets replaced by its value.',
```
ghost: Descriptions need to be translatable.

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+      ),
+    ),
+  );
+}
+
+/**
+ * Command callback: Execute specified PHP code.
```
ghost: The description after the colon should be an exact copy of the command description (see other commands for examples).

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+ */
+function php_eval_b_command($arguments, $options) {
+  try {
+    eval($arguments['code'] . ';');
```
ghost, line 35: If we're going to add a semicolon to the end of the PHP code, let's first make sure there isn't one already...

commands/eval.b.inc
```diff
@@ -0,0 +1,41 @@
...
+  }
+  catch(ParseError $e) {
+    // This is more readable than the default error we would get from PHP.
+    $err_msg = $e->getMessage() . ' ' . bt('in:') . ' ' . $arguments['code'];
```
ghost: I think the 'proper' way to write this would be:

```suggestion
    $err_msg = bt('@error in: @code', array(
      '@error' => $e->getMessage(),
      '@code' => $arguments['code'],
    ));
```

ghost (changes requested):

ghost:
I wonder if this command will even work as-is... I suspect `b` splits up arguments by whitespace, which means `b php-eval \'$node = node_load(1); print $node->title;\'` will turn out to be:

- `b`
- `php-eval` (command)
- `\'$node` (arg 1)
- `=` (arg 2)
- etc.

That's just a suspicion. I'd love to hear if you've actually tried this already and if it works @bugfolder. If not, we'll need to add `'multiple_argument' => 'code',` to the command descriptor and then try to merge multiple arguments from the same string together again, or (better yet) update `b` itself to allow for quoted arguments with whitespace...

commands/eval.b.inc
bugfolder: I copied the description and bootstrap level directly from the equivalent drush command, figuring that people moving from drush to b would expect to see the same description for the same commands that they were used to. But if your philosophy is to reword like this, I'm fine with the change.

commands/eval.b.inc
bugfolder: As with the description, I just followed the setting that drush used. I'm fine with removing it; can't think of any reason (other than the description) why full bootstrapping would be needed.

commands/eval.b.inc
bugfolder: Yes, there's a very good reason for the single quotes (and it's bitten me before). When you're doing a CLI command, the shell gets first crack at parsing before it passes the arguments to b, and if the shell sees a $ sign inside of a double-quoted string, it's going to try to substitute a value for that variable (just like PHP does), with likely unexpected results. If you use a double-quoted string, then b will never see "$node", it will see whatever the shell tries to substitute for $node (usually an empty string). 

So, if you try that command with the double-quoted version, it will fail (and in a very puzzling way).

commands/eval.b.inc
bugfolder: We don't need to. There is no harm in appending extra semicolons, even if superfluous. (That's also why the drush command does it this way as well.) And so it makes this code shorter and easier to understand.

bugfolder:
> I wonder if this command will even work as-is...

It does. I tried it during debugging and testing. It's worth going into a little detail on this, because b doesn't actually do the splitting on whitespace; it's the shell that parses the command and then passes the arguments to b. And while the shell does (usually) split on whitespace, within strings—both single-quoted and double-quoted—it includes whitespace within the string. So for both 
```
b php-eval '$node = node_load(1); print $node->title;'
```
and
```
b php-eval "$node = node_load(1); print $node->title;"
```
b receives just two arguments: the command and the string (even though the string contains whitespace).

The difference between the two is that in the second case, the shell will attempt to substitute for $node before passing the string to b, with the aforementioned puzzling results.

Specifically, the former will result in
```
Your first post!  
```
(that being what node 1 usually is titled), while the latter will result in
```
 [✘] syntax error, unexpected '=', expecting end of file in:  = node_load(1); print ->title; 
```
See how the "$node" got substituted away (by an empty string)? That's the difference.

So that's a definite "gotcha" waiting to happen to the unwary. Maybe there should be a warning in the help text for this command, although we can perhaps expect that people who are doing lots of command-line stuff already know about variable substitution in double-quoted text; if they're going to play with this dangerous command, they need to be prepared for what it does.

commands/eval.b.inc
bugfolder: OK, after some further testing, it is clear that `B_BOOTSTRAP_FULL` is needed.

A common use case for this command is to wrap a bunch of PHP commands that you'd like to execute inside of a function in some module, then you invoke that function with the command `b ev 'my_module_something()'`. For this to work, however, you need Backdrop fully bootstrapped, so that specifically the code that you added to your module gets included.

For that matter, even if you're just calling a Backdrop API function, you need Backdrop bootstrapped at least far enough to bring that function into scope. In general, for most real use cases, you're going to want things fully bootstrapped. And I'm sure the use cases where you need fully bootstrapped outnumber the cases where you'd want to use this before things are fully bootstrapped.

And besides, that's how drush does it. I realize that b aspires to be better than drush, but if people are coming to b from drush, they'll likely have the reasonable expectation that commands with the same name behave the same way.

commands/eval.b.inc
bugfolder: Per the next comment: since Backdrop is going to be fully bootstrapped, and that's an important property that users of this command will be relying on, it seems reasonable to mention it in the description.

bugfolder (commented):
On further scrutiny, the edit to the error message is not quite right: shouldn't use '@', should use '!' for the placeholders since output is not HTML.

commands/eval.b.inc
bugfolder: `bt()` wrappers added.

commands/eval.b.inc
bugfolder: Done.

commands/eval.b.inc
bugfolder: Let me also address the question of whether "(run/execute)" is needed. I am not particularly opposed to adding the parenthetical, but I do note that "evaluate" is the terminology used [in the PHP manual for this action](https://www.php.net/manual/en/function.eval).

commands/eval.b.inc
ghost: Ok, I can see your point. For posterity, here's what I tested (all commands are run from *outside* the Backdrop installation):

**Bootstrap required:**
- `b eval "echo 'Testing 123'"` -> ERROR: "The required bootstrap level for 'php-eval' is not ready."
- `b eval "echo 'Testing 123'" --root=backdrop` -> Outputs 'Testing 123' as expected
- `b eval "backdrop_cron_run()"` -> ERROR: "The required bootstrap level for 'php-eval' is not ready."
- `b eval "backdrop_cron_run()" --root=backdrop` -> Cron runs successfully (no output)

**Bootstrap *not* required:**
- `b eval "echo 'Testing 123'"` -> Outputs 'Testing 123' as expected
- `b eval "echo 'Testing 123'" --root=backdrop` -> Outputs 'Testing 123' as expected
- `b eval "backdrop_cron_run()"` -> ERROR: "PHP Fatal error:  Uncaught Error: Call to undefined function backdrop_cron_run() in /app/commands/eval.b.inc(34) : eval()'d code:1"
- `b eval "backdrop_cron_run()" --root=backdrop` -> ERROR: "PHP Fatal error:  Uncaught Error: Call to undefined function backdrop_cron_run() in /app/commands/eval.b.inc(34) : eval()'d code:1"

So while requiring bootstrap means you can't run commands outside of Backdrop, you can at least still run them all inside Backdrop (or by specifying the Backdrop installation using `--root`).

However if we remove the bootstrap requirement like I was suggesting, then it's impossible to call Backdrop functions either outside or inside a Backdrop install. Since that's a bigger problem, then I see the need for this now.

commands/eval.b.inc
ghost: As per my comment below, I agree that we need Backdrop bootstrapped for this command, so that text can stay in the description. However I still see the benefit of the bracketed text, so I'd like that added.

```suggestion
      'description' => bt('Evaluate (run/execute) arbitrary PHP code after bootstrapping Backdrop.'),
```

commands/eval.b.inc
ghost: Ok. So I had thought that might be the case, but when I looked at the third example, it has `$HOME` in double quotes, so I figured that'd be ok for the first example too. But I suppose `$HOME` is a bash variable, whereas `$node` is a PHP variable, and so they're handled differently...

commands/eval.b.inc
ghost: Thanks, but this'll still need to be updated when the description is updated with the bracketed text.

commands/eval.b.inc
ghost: I know "we don't need to", but that doesn't mean we shouldn't. Any empty PHP comment doesn't do any harm, but it wouldn't be allowed to remain in a core PR because we have coding standards and (as much as possible) we try to do things 'the right way'.

```suggestion
    $code = $arguments['code'];
    $code .= (substr($code, -1) != ';') ? ';' : '';
    eval($code);
```

ghost:
Thanks for the feedback re. whitespace. Good to know that it works as-is! Having said that, can you see any benefit to allowing multiple 'arguments' (i.e. multiple command strings). In other words, should we allow people to do this...?

`b eval "echo 'Command 1';" "echo 'Command 2';"`

If so, we can still add the `multiple_argument` key. But I'm guessing it's not necessary and people could instead just do this:

`b eval "echo 'Command 1'; echo 'Command 2';"`

bugfolder:
I see no strong reason for allowing multiple arguments. People can already execute multiple statements within a single argument as in your example, and it's simpler, and there's fewer quotes for them to keep track of.

commands/eval.b.inc
bugfolder: That's still going to add an extra ';' if there's trailing whitespace in the executed string because ';' would no longer be the final character, but adding `trim()` to line 34 would take care of that.

If the final character is a '}' then no ';' would be needed, but this routine would still add one. For example, this would be a valid input, no terminal semicolon needed: `if (TRUE) { print("foo"); }`.

So, yes, we could trap that one, too. But it seems like trying to trap all the cases where an extra semicolon would be superfluous adds needlessly to the code, even if those are the only two possibilities (whitespace and '}'). Just tacking on an extra semicolon covers all use cases, doesn't generate any extra PHP evaluation, and is more readable because it's shorter.

So I'll ask: is this really important?

commands/eval.b.inc
bugfolder: OK, will add.

commands/eval.b.inc
bugfolder: (run/execute) added to PR in both places.

commands/eval.b.inc
bugfolder: They're actually handled the same way; anything with a $ could be a bash variable and is treated as such inside a double-quoted string. So, for example, if I had in my bash script
```
node=foobar
b ev "print $node->title;"
```
then b would actually attempt to evaluate `'print foobar->title;'`

I could even do
```
node=\$user
b ev "print $node->name'"
```
and then b would actually evaluate `'print $user->name;'`

The moral being, I guess, stick with single quotes unless you're specifically after variable substitution.

But yes, $HOME is a predefined bash variable, so it will always get substituted.

commands/eval.b.inc
bugfolder: I'll also point out a few more examples that could be problematic for determining whether a terminal semicolon is needed:

* `b ev 'print "foo"; // some comment'`    <-- not needed
* `b ev 'print "foo" /* some comment */'`    <-- needed
* `b ev "print 'foo' // some comment ; "$'\n'" /* more */"`   <-- needed
* `b ev "print 'foo' // some comment "$'\n'" ; /* more */"`   <-- not needed

By the time we've got code that can handle the presence or absence of whitespace, braces, and comments, we're well down the path of parsing the PHP...before handing it over to the PHP parser.

Whereas tacking on an extra ';' does the job, and has years of successful use in drush. Just seems like we might as well stick with the proven, tested, and familiar approach.

ghost: Hmmm... I don't really like it, but fine. Let's leave it for now.

ghost:
Closing this in favour of the updated https://github.com/backdrop-contrib/bee/pull/143 PR.

PR #168, opened 2021-11-05
Issue #167: Fix install command's 'email' option.

### LINKED ISSUE #167: Running site-install in --auto mode uses admin@example.com instead of provided username.
I've been using this in my scripts:

`lando bee si --profile=standard --db=mysql://root:backdrop@localhost/backdrop --username=admin --password=password --site-name=my-site --auto`

I think it was working as expected till recently, but I tried it today, and could not log into the instance. I then used `bee uli` to log into the site as user1, only to find out that the username was set to `admin@example.com`.

ghost:
Fixes https://github.com/backdrop-contrib/bee/issues/167


klonos (approved):
LGTM 👍🏼

PR #175, opened 2022-03-04
Make context variable optional by adding a default value

ElusiveMind:
This is causing fatal errors when running on my environment which includes PHP 8.

### DIFF
--- a/bee.php
+++ b/bee.php
@@ -41,7 +41,7 @@ exit();
  *
  * @see https://www.php.net/manual/en/function.set-error-handler.php
  */
-function bee_error_handler($errno, $message, $filename, $line, $context) {
+function bee_error_handler($errno, $message, $filename, $line, $context = []) {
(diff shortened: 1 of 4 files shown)


bee.php
```diff
@@ -41,9 +41,19 @@
  *
  * @see https://www.php.net/manual/en/function.set-error-handler.php
  */
-function bee_error_handler($errno, $message, $filename, $line, $context) {
+function bee_error_handler($errno, $message, $filename, $line, $context = []) {
```
ghost, line 44: Backdrop (and Bee) still use the long array syntax, so this should be `array()` instead.

ghost (changes requested):
Since this PR is about the default value for `$context`, let's just fix that and remove all the other changes (they should be moved to a separate PR is necessary).

ghost:
Also, please rebase your code as I've fixed the issue with the CI tests since.

PR #182, opened 2022-04-08
Issue #181 - Fixes failure of database functions if special characters used

yorkshire-pudding:
Fixes #181 - Database functions fail if db_password contains special characters
Used the `rawurlencode` and `rawurldecode` that core uses

I've tested using the lando recipe in the module and they all pass (details below)  What I couldn't quite figure out is how to work with the `.lando.yml` and `setup.sh` script that calls the `install.sh` to test with a password that contains special characters

Perhaps this needs some dedicated tests writing?

I'll try and test on my hosting where I can more easily set a complex password than in lando

<details>
PHPUnit 8.5.26 #StandWithUkraine

Bee Core
 ✔ Bee is installed correctly
 ✔ Unknown command triggers an error
 ✔ Missing required argument triggers an error
 ✔ Commands can be called via an alias
 ✔ Root global option works
 ✔ Yes global option works
 ✔ Debug global option works

Cache Commands
 ✔ Cache clear command works

Config Commands
 ✔ Config get command works
 ✔ Config set command works
 ✔ Config export command works
 ✔ Config import command works

Cron Commands
 ✔ Cron command works

DBCommands
 ✔ Db export command works
 ✔ Db import command works

DBLog Commands
 ✔ Log command works

Download Commands
 ✔ Download command works
 ✔ Download core command works

Help Commands
 ✔ Help command works

PHPCommands
 ✔ Eval command works

Projects Commands
 ✔ Projects command works
 ✔ Enable command works
 ✔ Disable command works
 ✔ Uninstall command works
 ✔ Theme default command works
 ✔ Theme admin command works

Update Commands
 ✔ Update db command works

User Commands
 ✔ Users command works
 ✔ User password command works
 ✔ User login command works

Time: 43.26 seconds, Memory: 10.00 MB

OK (30 tests, 89 assertions)
PHPUnit 8.5.26 #StandWithUkraine

Multisite
 ✔ Site global option works

Multisite Install Commands
 ✔ Install command works

Multisite Download Commands
 ✔ Download command works

Time: 15.84 seconds, Memory: 10.00 MB

OK (3 tests, 20 assertions)
</details>

PR #183, opened 2022-04-09
Issue #181 - Fixes failure of database functions if special characters used

### LINKED ISSUE #181: Database functions fail if db_password contains special characters
This also applies to db_name and db_username but these are less likely to have special characters.
For the purposes of Backdrop where `rawurlencode` is used, special means anything except `A-Z, a-z, 0-1, - _ . ~`
https://www.php.net/manual/en/function.rawurlencode.php

**Steps to reproduce**

1. Create a backdrop database
2. Set the users password to have special characters
3. Attempt `install`, `db-import`, `db-export`, `sql`,  or `status --show-password`

**Expected result** works as normal

**Actual results** 
`install`, `db-import`, `db-export`, `sql` will all fail
`status --show-password` will show the encoded password

The backdrop GUI installer `core/install.php`, powered by `core/includes/install.inc`, uses `rawurlencode()` for db_name, db_username and db_password so what is stored in `settings.php` has these values encoded. 

If the `rawurldecode()` function is not used to read the database information stored in `settings.php` then the value returned will be encoded and so won't match the actual DB password.

yorkshire-pudding:
Fixes #181 - Database functions fail if db_password contains special characters
Used the `rawurlencode` and `rawurldecode` that core uses

I've tested using the lando recipe in the module and they all pass (details below).
<details>
PHPUnit 8.5.26 #StandWithUkraine

Bee Core
 ✔ Bee is installed correctly
 ✔ Unknown command triggers an error
 ✔ Missing required argument triggers an error
 ✔ Commands can be called via an alias
 ✔ Root global option works
 ✔ Yes global option works
 ✔ Debug global option works

Cache Commands
 ✔ Cache clear command works

Config Commands
 ✔ Config get command works
 ✔ Config set command works
 ✔ Config export command works
 ✔ Config import command works

Cron Commands
 ✔ Cron command works

DBCommands
 ✔ Db export command works
 ✔ Db import command works

DBLog Commands
 ✔ Log command works

Download Commands
 ✔ Download command works
 ✔ Download core command works

Help Commands
 ✔ Help command works

PHPCommands
 ✔ Eval command works

Projects Commands
 ✔ Projects command works
 ✔ Enable command works
 ✔ Disable command works
 ✔ Uninstall command works
 ✔ Theme default command works
 ✔ Theme admin command works

Update Commands
 ✔ Update db command works

User Commands
 ✔ Users command works
 ✔ User password command works
 ✔ User login command works

Time: 43.26 seconds, Memory: 10.00 MB

OK (30 tests, 89 assertions)
PHPUnit 8.5.26 #StandWithUkraine

Multisite
 ✔ Site global option works

Multisite Install Commands
 ✔ Install command works

Multisite Download Commands
 ✔ Download command works

Time: 15.84 seconds, Memory: 10.00 MB

OK (3 tests, 20 assertions)
</details>
What I couldn't quite figure out is how to work with the `.lando.yml` and `setup.sh` script that calls the `install.sh` to test with a password that contains special characters

Perhaps this needs some dedicated tests writing?

I'll try and test on my hosting where I can more easily set a complex password than in lando

### DIFF
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -136,7 +136,7 @@ function sql_bee_callback($arguments, $options) {
  */
 function db_bee_mysql_options($db_info) {
 
-  $options .= ' --user=' . rawurldecode($db_info['username']);
+  $options = ' --user=' . rawurldecode($db_info['username']);
   $options .= " --password='" .  rawurldecode($db_info['password']) . "'";
   $options .= ' --host=' . $db_info['host'];
   $options .= ' ' . rawurldecode($db_info['database']);
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -135,8 +135,7 @@ function sql_bee_callback($arguments, $options) {
  *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
  */
 function db_bee_mysql_options($db_info) {
-
-  $options = ' --user=' . rawurldecode($db_info['username']);
+  $options = '--user=' . rawurldecode($db_info['username']);
   $options .= " --password='" .  rawurldecode($db_info['password']) . "'";
   $options .= ' --host=' . $db_info['host'];
   $options .= ' ' . rawurldecode($db_info['database']);
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -133,7 +133,7 @@ function bee_compile_mysql_command($mysql_function = 'mysql') {
   // Define command line base
   $command =  $mysql_function;
   $command .= ' --user=' . rawurldecode($db_info['username']);
-  $command .= ' --password=' . rawurldecode($db_info['password']);
+  $command .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
   $command .= ' --host=' . $db_info['host'];
   $command .= ' ' . rawurldecode($db_info['database']);
 
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -63,7 +63,7 @@ function db_export_bee_callback($arguments, $options) {
   $filename = isset($arguments['file']) ? $arguments['file'] : date('Ymd_His') . '.sql';
 
   // Export and compress the database.
-  $export_command = bee_compile_mysql_command('mysqldump') . ' | gzip > ' . $filename . '.gz';
+  $export_command = 'mysqldump ' . db_bee_mysql_options($db_info) . ' | gzip > ' . $filename . '.gz';
   exec($export_command, $output, $result);
 
   if ($result === 0) {
@@ -96,7 +96,7 @@ function db_import_bee_callback($arguments, $options) {
   if ($gzip) {
     $import_command .= "gunzip -c $filename | ";
   }
-  $import_command .= bee_compile_mysql_command();
+  $import_command .= 'mysql ' . db_bee_mysql_options($db_info);
   if (!$gzip) {
     $import_command .= " < $filename";
   }
@@ -115,28 +115,32 @@ function db_import_bee_callback($arguments, $options) {
  * database credentials.
  */
 function sql_bee_callback($arguments, $options) {
-
-  // Open SQL command-line.
-  $command = bee_compile_mysql_command();
-  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
-}
-
-/**
- * Helper function to create the mysql command line
- * using database credentials.
- */
-function bee_compile_mysql_command($mysql_function = 'mysql') {
   // Get database info.
   $db_connection = Database::getConnectionInfo();
   $db_info = $db_connection['default'];
 
-  // Define command line base
-  $command =  $mysql_function;
-  $command .= ' --user=' . rawurldecode($db_info['username']);
-  $command .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
-  $command .= ' --host=' . $db_info['host'];
-  $command .= ' ' . rawurldecode($db_info['database']);
+  // Open SQL command-line.
+  $command = 'mysql ' . db_bee_mysql_options($db_info);
+  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+}
 
-  // Return the concatenated command
-  return $command;
+/**
+ * Returns a decoded options string for use with MySQL commands.
+ *
+ * @param array $db_info
+ *   Connection information for the current database. Generally obtained via
+ *   Database::getConnectionInfo().
+ *
+ * @return string
+ *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
+ */
+function db_bee_mysql_options($db_info) {
+
+  $options =  $mysql_function;
+  $options .= ' --user=' . rawurldecode($db_info['username']);
+  $options .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
+  $options .= ' --host=' . $db_info['host'];
+  $options .= ' ' . rawurldecode($db_info['database']);
+
+  return $options;
 }
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -138,6 +138,5 @@ function bee_compile_mysql_command($mysql_function = 'mysql') {
   $command .= ' ' . rawurldecode($db_info['database']);
 
   // Return the concatenated command
-  echo $command;
   return $command;
 }
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -136,9 +136,8 @@ function sql_bee_callback($arguments, $options) {
  */
 function db_bee_mysql_options($db_info) {
 
-  $options =  $mysql_function;
   $options .= ' --user=' . rawurldecode($db_info['username']);
-  $options .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
+  $options .= " --password='" .  rawurldecode($db_info['password']) . "'";
   $options .= ' --host=' . $db_info['host'];
   $options .= ' ' . rawurldecode($db_info['database']);
 
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -63,12 +63,12 @@ function db_export_bee_callback($arguments, $options) {
   $filename = isset($arguments['file']) ? $arguments['file'] : date('Ymd_His') . '.sql';
 
   // Export and compress the database.
-  $export_command = 'mysqldump --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'] . ' | gzip > ' . $filename . '.gz';
+  $export_command = bee_compile_mysql_command('mysqldump') . ' | gzip > ' . $filename . '.gz';
   exec($export_command, $output, $result);
 
   if ($result === 0) {
     bee_message(bt("The '!database' database was exported to '!filename.gz'.", array(
-      '!database' => $db_info['database'],
+      '!database' => rawurldecode($db_info['database']),
       '!filename' => $_bee_backdrop_root . '/' . $filename,
     )), 'success');
   }
@@ -96,7 +96,7 @@ function db_import_bee_callback($arguments, $options) {
   if ($gzip) {
     $import_command .= "gunzip -c $filename | ";
   }
-  $import_command .= 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'];
+  $import_command .= bee_compile_mysql_command();
   if (!$gzip) {
     $import_command .= " < $filename";
   }
@@ -105,7 +105,7 @@ function db_import_bee_callback($arguments, $options) {
   if ($result === 0) {
     bee_message(bt("'!filename' was imported into the '!database' database.", array(
       '!filename' => $filename,
-      '!database' => $db_info['database'],
+      '!database' => rawurldecode($db_info['database']),
     )), 'success');
   }
 }
@@ -115,11 +115,29 @@ function db_import_bee_callback($arguments, $options) {
  * database credentials.
  */
 function sql_bee_callback($arguments, $options) {
+
+  // Open SQL command-line.
+  $command = bee_compile_mysql_command();
+  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+}
+
+/**
+ * Helper function to create the mysql command line
+ * using database credentials.
+ */
+function bee_compile_mysql_command($mysql_function = 'mysql') {
   // Get database info.
   $db_connection = Database::getConnectionInfo();
   $db_info = $db_connection['default'];
 
-  // Open SQL command-line.
-  $command = 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'];
-  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+  // Define command line base
+  $command =  $mysql_function;
+  $command .= ' --user=' . rawurldecode($db_info['username']);
+  $command .= ' --password=' . rawurldecode($db_info['password']);
+  $command .= ' --host=' . $db_info['host'];
+  $command .= ' ' . rawurldecode($db_info['database']);
+
+  // Return the concatenated command
+  echo $command;
+  return $command;
 }
--- a/commands/install.bee.inc
+++ b/commands/install.bee.inc
@@ -71,7 +71,7 @@ function install_bee_command() {
       'aliases' => array('si', 'site-install'),
       'examples' => array(
         'bee install' => bt('Install Backdrop in interactive mode, providing information when prompted.'),
-        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --db-host=db_server --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
         'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
       ),
     ),
@@ -99,6 +99,7 @@ function install_bee_callback($arguments, $options) {
   // Get database name.
   if (!empty($options['db-name'])) {
     $db_name = $options['db-name'];
+    $check_db_name_option = TRUE;
   }
   else {
     $db_name = rawurlencode(bee_check_plain(bee_input(bt('Database name'), '', TRUE)));
@@ -125,7 +126,7 @@ function install_bee_callback($arguments, $options) {
     $db_host = $options['db-host'];
   }
   else {
-    $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
+    $check_db_name_option ? $db_host = 'localhost' : $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
   }
 
   // Compile database url
--- a/commands/install.bee.inc
+++ b/commands/install.bee.inc
@@ -13,9 +13,21 @@ function install_bee_command() {
       'description' => bt('Install Backdrop and setup a new site.'),
       'callback' => 'install_bee_callback',
       'options' => array(
-        'db' => array(
-          'description' => bt("A database connection string of the form: 'mysql://user:pass@localhost/database_name'. See https://api.backdropcms.org/database-configuration for more details."),
-          'value' => bt('String'),
+        'db-name' => array(
+          'description' => bt("The database name."),
+          'value' => bt('Database name'),
+        ),
+        'db-user' => array(
+          'description' => bt("The username for the database."),
+          'value' => bt('Database username'),
+        ),
+        'db-pass' => array(
+          'description' => bt("The password for the database. "),
+          'value' => bt('Database password'),
+        ),
+        'db-host' => array(
+          'description' => bt("The host for the database relative to your webserver. Defaults to 'localhost'"),
+          'value' => bt('Database host'),
         ),
         'username' => array(
           'description' => bt("The username for the primary administrator account. Defaults to 'admin'."),
@@ -59,8 +71,8 @@ function install_bee_command() {
       'aliases' => array('si', 'site-install'),
       'examples' => array(
         'bee install' => bt('Install Backdrop in interactive mode, providing information when prompted.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
       ),
     ),
   );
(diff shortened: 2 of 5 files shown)


yorkshire-pudding:
Tested on host.
Works with complex password in interactive mode but not in auto mode.  I think this is because the `parse_url` function doesn't work with characters that are invalid as part of a URL, whereas with interactive we encode each element.
I came up with a different idea initially but thought `parse_url` was more efficient.  I will revisit that previous idea as we want it to work with secure passwords in auto mode.
The other option is a breaking change and splitting the constituent parts of the db option into individual options so they can be encoded individually.

yorkshire-pudding:
Tested this with the lando recipe
<details>
PHPUnit 8.5.26 #StandWithUkraine

Bee Core
 ✔ Bee is installed correctly
 ✔ Unknown command triggers an error
 ✔ Missing required argument triggers an error
 ✔ Commands can be called via an alias
 ✔ Root global option works
 ✔ Yes global option works
 ✔ Debug global option works

Cache Commands
 ✔ Cache clear command works

Config Commands
 ✔ Config get command works
 ✔ Config set command works
 ✔ Config export command works
 ✔ Config import command works

Cron Commands
 ✔ Cron command works

DBCommands
 ✔ Db export command works
 ✔ Db import command works

DBLog Commands
 ✔ Log command works

Download Commands
 ✔ Download command works
 ✔ Download core command works

Help Commands
 ✔ Help command works

PHPCommands
 ✔ Eval command works

Projects Commands
 ✔ Projects command works
 ✔ Enable command works
 ✔ Disable command works
 ✔ Uninstall command works
 ✔ Theme default command works
 ✔ Theme admin command works

Update Commands
 ✔ Update db command works

User Commands
 ✔ Users command works
 ✔ User password command works
 ✔ User login command works

Time: 49.68 seconds, Memory: 10.00 MB

OK (30 tests, 89 assertions)
PHPUnit 8.5.26 #StandWithUkraine

Multisite
 ✔ Site global option works

Multisite Install Commands
 ✔ Install command works

Multisite Download Commands
 ✔ Download command works

Time: 16.71 seconds, Memory: 10.00 MB

OK (3 tests, 20 assertions)
</details>

Also tested on host and this now works with complex password both auto and interactively.
The only (unlikely) condition I can see tripping this up, and would be tricky to accommodate is if someone put a `:` character in their database username.

commands/db.bee.inc
```diff
@@ -115,11 +115,28 @@ function db_import_bee_callback($arguments, $options) {
...
+  // Open SQL command-line.
+  $command = bee_compile_mysql_command();
+  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+}
+
+/**
+ * Helper function to create the mysql command line
+ * using database credentials.
+ */
+function bee_compile_mysql_command($mysql_function = 'mysql') {
```
ghost: Let's simplify this function a bit - make it just handle the options part of the MySQL commands, and let's pass in the Db connection info, since we're usually getting it before calling this anyway.

```suggestion
/**
 * Returns a decoded options string for use with MySQL commands.
 *
 * @param array $db_info
 *   Connection information for the current database. Generally obtained via
 *   Database::getConnectionInfo().
 *
 * @return string
 *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
 */
function db_bee_mysql_options($db_info) {
```

commands/db.bee.inc
```diff
@@ -115,11 +115,28 @@ function db_import_bee_callback($arguments, $options) {
...
   // Get database info.
   $db_connection = Database::getConnectionInfo();
   $db_info = $db_connection['default'];
```
ghost: Remove.

commands/db.bee.inc
```diff
@@ -115,11 +115,28 @@ function db_import_bee_callback($arguments, $options) {
...
-  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+  // Define command line base
+  $command =  $mysql_function;
+  $command .= ' --user=' . rawurldecode($db_info['username']);
+  $command .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
+  $command .= ' --host=' . $db_info['host'];
+  $command .= ' ' . rawurldecode($db_info['database']);
+
+  // Return the concatenated command
+  return $command;
```
ghost:
```suggestion
  $options = '--user=' . rawurldecode($db_info['username']);
  $options .= ' --password=' . rawurldecode($db_info['password']);
  $options .= ' --host=' . $db_info['host'];
  $options .= ' ' . rawurldecode($db_info['database']);

  return $options;
```

commands/db.bee.inc
```diff
@@ -63,12 +63,12 @@ function db_export_bee_callback($arguments, $options) {
   $filename = isset($arguments['file']) ? $arguments['file'] : date('Ymd_His') . '.sql';
 
   // Export and compress the database.
-  $export_command = 'mysqldump --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'] . ' | gzip > ' . $filename . '.gz';
+  $export_command = bee_compile_mysql_command('mysqldump') . ' | gzip > ' . $filename . '.gz';
```
ghost:
```suggestion
  $export_command = 'mysqldump ' . db_bee_mysql_options($db_info) . ' | gzip > ' . $filename . '.gz';
```

commands/db.bee.inc
```diff
@@ -96,7 +96,7 @@ function db_import_bee_callback($arguments, $options) {
   if ($gzip) {
     $import_command .= "gunzip -c $filename | ";
   }
-  $import_command .= 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'];
+  $import_command .= bee_compile_mysql_command();
```
ghost:
```suggestion
  $import_command .= 'mysql ' . db_bee_mysql_options($db_info);
```

commands/db.bee.inc
```diff
@@ -115,11 +115,28 @@ function db_import_bee_callback($arguments, $options) {
  * database credentials.
  */
 function sql_bee_callback($arguments, $options) {
+
+  // Open SQL command-line.
+  $command = bee_compile_mysql_command();
+  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
```
ghost:
```suggestion
  // Get database info.
  $db_connection = Database::getConnectionInfo();
  $db_info = $db_connection['default'];
  
  // Open SQL command-line.
  $command = 'mysql ' . db_bee_mysql_options($db_info);
  proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
```

commands/install.bee.inc
```diff
@@ -86,12 +86,19 @@ function install_bee_callback($arguments, $options) {
...
   if (!empty($options['db'])) {
-    $db = $options['db'];
+    // Split the passed URL into parts so can encode user, password and db name
+    $db_url = $options['db'];
+    $db_name = rawurlencode(bee_check_plain(bee_after_last('/', $db_url)));
+    $db_username = rawurlencode(bee_check_plain(bee_before(':', bee_after('mysql://', $db_url))));
+    $db_password = rawurlencode(bee_check_plain(bee_between($db_username . ':', '@' . bee_after_last('@', $db_url), $db_url)));
+    $db_host = bee_check_plain(bee_between_last('@', '/', $db_url));
+    // Re-compile the database URL using the sanitized parts
+    $db = "mysql://$db_username:$db_password@$db_host/$db_name";
```
ghost: Please update this whole file's changes to instead do something similar to what `install.sh` does: https://github.com/backdrop/backdrop/blob/8762dc9369529b6238dda03318394d0f46dc6640/core/scripts/install.sh#L124

Using `parse_url()` should avoid the need for all the helper functions, and `array_map('urldecode', ...)` should simplify things too.

ghost (changes requested):
Thanks @yorkshire-pudding. It took me a while to get my head around all this. Mainly because I didn't notice at first the different places where you used `rawurlencode()` and `rawurldecode()` (I thought they were all decoding, and that threw me for a while). But finally worked it out and now left some comments/requested changes for you.

commands/install.bee.inc
yorkshire-pudding: @BWPanda - I did try using this, but `parse_url()` is designed for standard hyperlinks and whenever I passed the connection string whole (`--auto` rather than interactive) it failed.  It will not work on a database connection URL with complex passwords.  Take for example `xYK1jxu3$RSOEmL#evqI`; it cannot parse it.  Both the `$` and the `#` cause problems and you can't quote them as that's not valid in a URL either.  I ultimately think `install.sh` is flawed in that it will only accept simple MySQL passwords, whereas MySQL will accept complex passwords and also the GUI installer will handle complex passwords because it takes each element individually.

See https://github.com/backdrop-contrib/bee/pull/183#issuecomment-1094071869

[See where it was initially included](https://github.com/backdrop-contrib/bee/pull/182/commits/e8f96881b3834408117a507064f112cff85c5694#diff-29f05d9cc78b79338561837d2f67cb48fc7cb56a644bed66ca8477208a66ed3e)

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: @BWPanda - please can we discuss?

ghost: Yeah, looks like core still suffers from the same problem: https://github.com/backdrop/backdrop-issues/issues/2904

I think the easiest solution then is to ditch the `db` option and just provide options for each component individually...

yorkshire-pudding:
@BWPanda - As per our discussion, I've changed the code to split the database settings.  I've also changed the test (multi-site) where it uses this command.

I've run tests using the updated lando recipe:

<details>
PHPUnit 8.5.26 #StandWithUkraine

Bee Core
 ✔ Bee is installed correctly
 ✔ Unknown command triggers an error
 ✔ Missing required argument triggers an error
 ✔ Commands can be called via an alias
 ✔ Root global option works
 ✔ Yes global option works
 ✔ Debug global option works

Cache Commands
 ✔ Cache clear command works

Config Commands
 ✔ Config get command works
 ✔ Config set command works
 ✔ Config export command works
 ✔ Config import command works

Cron Commands
 ✔ Cron command works

DBCommands
 ✔ Db export command works
 ✔ Db import command works

DBLog Commands
 ✔ Log command works

Download Commands
 ✔ Download command works
 ✔ Download core command works

Help Commands
 ✔ Help command works

PHPCommands
 ✔ Eval command works

Projects Commands
 ✔ Projects command works
 ✔ Enable command works
 ✔ Disable command works
 ✔ Uninstall command works
 ✔ Theme default command works
 ✔ Theme admin command works

Update Commands
 ✔ Update db command works

User Commands
 ✔ Users command works
 ✔ User password command works
 ✔ User login command works

Time: 41.96 seconds, Memory: 10.00 MB

OK (30 tests, 89 assertions)
PHPUnit 8.5.26 #StandWithUkraine

Multisite
 ✔ Site global option works

Multisite Install Commands
 ✔ Install command works

Multisite Download Commands
 ✔ Download command works

Time: 15.69 seconds, Memory: 10.00 MB

OK (3 tests, 20 assertions)
</details>

commands/install.bee.inc
yorkshire-pudding: Done - see comment below.

yorkshire-pudding:
@BWPanda - please can you review the latest commit

ghost:
It's on my to do list.

commands/db.bee.inc
```diff
@@ -120,6 +120,27 @@ function sql_bee_callback($arguments, $options) {
...
+ * @param array $db_info
+ *   Connection information for the current database. Generally obtained via
+ *   Database::getConnectionInfo().
+ *
+ * @return string
+ *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
+ */
+function db_bee_mysql_options($db_info) {
+
+  $options =  $mysql_function;
```
ghost: This needs to go.

commands/db.bee.inc
```diff
@@ -120,6 +120,27 @@ function sql_bee_callback($arguments, $options) {
...
+  $options .= ' --user=' . rawurldecode($db_info['username']);
+  $options .= ' --password=\'' .  rawurldecode($db_info['password']) . '\'' ;
```
ghost: Rather than having escaped characters, better to just use different quotes:

```suggestion
  $options .= " --password='" . rawurldecode($db_info['password']) . "'";
```

commands/install.bee.inc
```diff
@@ -59,8 +71,8 @@ function install_bee_command() {
       'aliases' => array('si', 'site-install'),
       'examples' => array(
         'bee install' => bt('Install Backdrop in interactive mode, providing information when prompted.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
```
ghost: Neither of these examples mention the host parameter...

commands/install.bee.inc
```diff
@@ -84,17 +96,40 @@ function install_bee_callback($arguments, $options) {
...
+  if (!empty($options['db-host'])) {
+    $db_host = $options['db-host'];
   }
   else {
-    $db_name = bee_check_plain(bee_input(bt('Database name'), '', TRUE));
-    $db_username = bee_check_plain(bee_input(bt('Database username'), '', TRUE));
-    $db_password = bee_check_plain(bee_input(bt('Database password'), '', TRUE));
     $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
-    $db = "mysql://$db_username:$db_password@$db_host/$db_name";
   }
```
ghost: With the host, I think it'd be nice to just assume it's `localhost` when the user provides a db-name (i.e. they've provided info via the command (meaning they probably don't want to be prompted), but they haven't provided the host so it's likely just the default).

ghost (changes requested):
A few things to change please.

commands/db.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done

yorkshire-pudding:
Thanks @BWPanda for your feedback.  I've implemented these.

commands/install.bee.inc
yorkshire-pudding: Done

commands/db.bee.inc
```diff
@@ -120,6 +120,26 @@ function sql_bee_callback($arguments, $options) {
...
+ * @param array $db_info
+ *   Connection information for the current database. Generally obtained via
+ *   Database::getConnectionInfo().
+ *
+ * @return string
+ *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
+ */
+function db_bee_mysql_options($db_info) {
+
+  $options .= ' --user=' . rawurldecode($db_info['username']);
```
ghost:
```suggestion
  $options = '--user=' . rawurldecode($db_info['username']);
```

commands/install.bee.inc
```diff
@@ -84,17 +96,41 @@ function install_bee_callback($arguments, $options) {
     bee_render_text(array('value' => bt("Running installation in interactive mode. To run automatically (i.e. non-interactively), use the '--auto' option.\n")));
   }
 
-  // Get database connection string.
-  if (!empty($options['db'])) {
-    $db = $options['db'];
+  // Get database name.
+  if (!empty($options['db-name'])) {
+    $db_name = $options['db-name'];
+    $check_db_name_option = TRUE;
```
ghost:
```suggestion
```

commands/install.bee.inc
```diff
@@ -84,17 +96,41 @@ function install_bee_callback($arguments, $options) {
...
+    $db_password = rawurlencode(bee_check_plain(bee_input(bt('Database password'), '', TRUE)));
+  }
+
+  // Get database host.
+  if (!empty($options['db-host'])) {
+    $db_host = $options['db-host'];
+  }
+  else {
+    $check_db_name_option ? $db_host = 'localhost' : $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
+  }
```
ghost:
```suggestion
  elseif (!empty($options['db-name'])) {
    // If the user provided a DB name but not the host, we assume 'localhost'.
    $db_host = 'localhost';
  }
  else {
    $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
  }
```

ghost (changes requested):

commands/db.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done.

commands/install.bee.inc
yorkshire-pudding: Done.

yorkshire-pudding:
@BWPanda - Done those.  

Regarding checking the db name, are there any general principles about when you would test once and store in a variable versus checking the same test each time?  I'm learning plenty, but sometimes I can't see why one approach would be favoured over another and would like to understand more.

ghost:
@yorkshire-pudding I understand where you're coming from. If it were performing a DB query or reading from config files, then yes, doing that once and then storing the value in a local variable would be better. But in this case, checking the DB name is just checking a variable (albeit an array rather than a string), so there's no real benefit storing this value in a second variable for later. Hope that makes sense...?

yorkshire-pudding:
Thanks. That's a good explanation. Makes sense.

commands/db.bee.inc
```diff
@@ -120,6 +120,26 @@ function sql_bee_callback($arguments, $options) {
...
+ * @param array $db_info
+ *   Connection information for the current database. Generally obtained via
+ *   Database::getConnectionInfo().
+ *
+ * @return string
+ *   A string of command options that can be appended to 'mysql' or 'mysqldump'.
+ */
+function db_bee_mysql_options($db_info) {
+
+  $options = ' --user=' . rawurldecode($db_info['username']);
```
ghost: Please remove this leading blank line and the leading space before the `--user` (that's already accounted for when we call `db_bee_mysql_options()` above).

```suggestion
  $options = '--user=' . rawurldecode($db_info['username']);
```

commands/install.bee.inc
```diff
@@ -13,9 +13,21 @@ function install_bee_command() {
       'description' => bt('Install Backdrop and setup a new site.'),
       'callback' => 'install_bee_callback',
       'options' => array(
-        'db' => array(
-          'description' => bt("A database connection string of the form: 'mysql://user:pass@localhost/database_name'. See https://api.backdropcms.org/database-configuration for more details."),
-          'value' => bt('String'),
+        'db-name' => array(
+          'description' => bt("The database name."),
```
ghost:
```suggestion
          'description' => bt("The name of the database to install into."),
```

commands/install.bee.inc
```diff
@@ -13,9 +13,21 @@ function install_bee_command() {
...
+          'value' => bt('Database name'),
+        ),
+        'db-user' => array(
+          'description' => bt("The username for the database."),
```
ghost:
```suggestion
          'description' => bt("The username for connecting to the database."),
```

commands/install.bee.inc
```diff
@@ -13,9 +13,21 @@ function install_bee_command() {
...
+          'value' => bt('Database username'),
+        ),
+        'db-pass' => array(
+          'description' => bt("The password for the database. "),
```
ghost:
```suggestion
          'description' => bt("The password for connecting to the database."),
```

commands/install.bee.inc
```diff
@@ -13,9 +13,21 @@ function install_bee_command() {
...
+          'value' => bt('Database password'),
+        ),
+        'db-host' => array(
+          'description' => bt("The host for the database relative to your webserver. Defaults to 'localhost'"),
```
ghost: I think we should mention the fact that specifying a DB name but not host causes this to default. Also not sure what "relative to your webserver" means ('relative' makes it  sound like a path)...

```suggestion
          'description' => bt("The host for the database. Defaults to 'localhost' when a database name is provided."),
```

commands/install.bee.inc
```diff
@@ -59,8 +71,8 @@ function install_bee_command() {
       'aliases' => array('si', 'site-install'),
       'examples' => array(
         'bee install' => bt('Install Backdrop in interactive mode, providing information when prompted.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
-        'bee install --db=mysql://admin:P@ssw0rd!@localhost/backdrop --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --db-host=db_server --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
+        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
```
ghost: I think the host in these examples should be switched, since the first examples gives the bare minimum and leaves the default values for the rest.

```suggestion
        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --auto' => bt('Install Backdrop automatically using the provided database credentials, and default settings for everything else.'),
        'bee install --db-name=backdrop --db-user=admin --db-password=P@ssw0rd! --db-host=db_server --username=Root --password=N0tS3cur3 --email=root@mydomain.com --site-name="My awesome site!"' => bt('Install Backdrop using the given options, and be prompted for the rest.'),
```

commands/install.bee.inc
```diff
@@ -84,17 +96,44 @@ function install_bee_callback($arguments, $options) {
...
+  elseif (!empty($options['db-name'])) {
+    // If the user provided a DB name but not the host, we assume 'localhost'.
+    $db_host = 'localhost';
+  }
+  else { 
     $db_host = bee_check_plain(bee_input(bt('Database host'), 'localhost', TRUE));
-    $db = "mysql://$db_username:$db_password@$db_host/$db_name";
   }
+
+  // Compile database url
```
ghost: Comments should end with a full-stop.

```suggestion
  // Compile database URL.
```

ghost (changes requested):
Sorry, looking more closely I found a few more things to change.

commands/db.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done.

commands/install.bee.inc
yorkshire-pudding: Done

commands/install.bee.inc
yorkshire-pudding: Done.

yorkshire-pudding:
Hi @BWPanda - I have done those.

ghost (approved):

PR #184, opened 2022-04-25
Issue #22: Provides a verbose option that prevents bee from outputting warnings…

### LINKED ISSUE #22: Add a --quiet option for bee commands.

rbargerhuff:
Fixes #22

### DIFF
--- a/includes/globals.inc
+++ b/includes/globals.inc
@@ -24,6 +24,10 @@ $_bee_global_options = array(
     'description' => bt("Enables 'debug' mode, in which 'log' type messages will be displayed (in addition to all other messages)."),
     'short' => 'd',
   ),
+  'verbose' => array(
+    'description' => bt("Enables the output of PHP notices and warnings."),
+    'short' => 'v',
+  ),
 );
 
 // The command that was given (e.g. `core-status`).
@@ -39,7 +43,7 @@ $_bee_options = array();
 // A list that maps command aliases to their commands.
 $_bee_command_aliases = array();
 
-// A boolean indicating whether or not Backdrop is installed.
+// A boolean indicating whether Backdrop is installed.
 $_bee_backdrop_installed = FALSE;
 
 // The root directory of the Backdrop installation to use.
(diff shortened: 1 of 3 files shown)


includes/globals.inc
```diff
@@ -51,12 +55,15 @@ $_bee_backdrop_site = FALSE;
...
 
-// A boolean indicating whether or not `Yes` mode is enabled.
+// A boolean indicating whether `Yes` mode is enabled.
 $_bee_yes_mode = FALSE;
 
-// A boolean indicating whether or not `Debug` mode is enabled.
+// A boolean indicating whether `Debug` mode is enabled.
 $_bee_debug_mode = FALSE;
 
+// A boolean indicating whether 'verbose' mode is enabled.
```
ghost: Best to copy syntax from other examples above:

```suggestion
// A boolean indicating whether `Verbose` mode is enabled.
```

ghost (changes requested):
One very small fix.

rbargerhuff:
My collegue and I who originally made this Pull Request noticed that the change made by @BWPanda had not been commited to the main branch and we are wondering how this request can be included.

rbargerhuff (commented):
This change is great. Thank you!

yorkshire-pudding:
Hi @rbargerhuff - I'm sorry this wasn't merged earlier.  Would you be willing to to a rebase of your PR so that it can be merged as it is currently causing a conflict due to recent changes for PHP 8.1 compatibility?
In particular, this line in `bee.php`
```php
function bee_error_handler($errno, $message, $filename, $line, $context) {
```
has changed to
```php
function bee_error_handler($errno, $message, $filename, $line, $context = NULL) {
```

And because your branch is based on an older version, there is a conflict there.
This looks like a great addition so would be great to get it in.

rbargerhuff:
Abandoning this PR since it is no longer applicable.

Issue #195, opened 2022-08-01
Allow setting HTTP_HOST directly via --uri option


quicksketch:
I'm using bee together with ddev and finding that bee can't figure out what the hostname should be.

I've set up my ddev command as follows:
```
/usr/local/bin/bee/bee.php --root=/var/www/html/$DDEV_DOCROOT $@
```

When using a ddev command like `ddev bee uli`, I get back the output:

```
Use the following link to login as 'admin':
http://web/user/reset/1/1659380860/sA-ge2y4h7gkLsH4TEvAOSxGGeE_OlQaz4qBWRsccy0/login
```

The problem is bee doesn't know the URI of my site and it uses the hostname from the container. I can "fix" this by setting a `$base_url` in `settings.php`, but I would find it preferable if I could set the URI as an option.

I tried using `--site=$DDEV_PRIMARY_URL`, but the `--site` option is intended to select a multisite, and does not do much of anything when used on a single-site installation.

Proposed solution:

* Support a `--uri` option similar to Drush: https://docs.drush.org/en/8.x/usage/
* This would enable setting the URI with `/usr/local/bin/bee/bee.php --uri=$DDEV_PRIMARY_URL`

quicksketch:
I filed a PR at https://github.com/backdrop-contrib/bee/pull/196 that adds the option and documentation. I did not add any validation of the new option; do we validate other global options somewhere?

yorkshire-pudding:
@quicksketch 
I've confirmed that this issue also affects Lando.
I've tested in Lando (using the bee lando recipe) and while it does return the correct link, it also produces quite  a few error messages:

<details>

```
user@machine-name:~/projects/backdrop/bee-195-uri/bee$ lando bee --root=backdrop --uri=https://bee.lndo.site uli
Undefined index: port
 /app/includes/miscellaneous.inc:54

ini_set(): Headers already sent. You cannot change the session module's ini settings at this time
 /app/backdrop/settings.php:168

ini_set(): Headers already sent. You cannot change the session module's ini settings at this time
 /app/backdrop/settings.php:169

ini_set(): Headers already sent. You cannot change the session module's ini settings at this time
 /app/backdrop/settings.php:177

ini_set(): Headers already sent. You cannot change the session module's ini settings at this time
 /app/backdrop/settings.php:184

ini_set(): Headers already sent. You cannot change the session module's ini settings at this time
 /app/backdrop/core/includes/bootstrap.inc:979

[...]

PR #196, opened 2022-08-01
Issue #195: Add support for --uri option.

quicksketch:
Initial work to support --uri option: https://github.com/backdrop-contrib/bee/issues/195

To test:

* Do not specify a $base_url in settings.php
* Use a command that generates a URL, such as `uli`: `bee --uri=https://example.com uli`
* The output should use the specified URI as the base URL, e.g. https://example.com/user/reset/1/1659390793/[hash]/login

Using https vs. http, sub-directories, and port numbers should all work.

### DIFF
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -49,7 +49,20 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
+
+    
+    if ($_bee_options['uri']) {
+      $uri_parts = parse_url($_bee_options['uri']);
+      $_SERVER['HTTP_HOST'] = $uri_parts['host'] . ($uri_parts['port'] ? ':' . $uri_parts['port'] : '');
+      $_SERVER['HTTPS'] = $uri_parts['scheme'] === 'https';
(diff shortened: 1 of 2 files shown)


includes/miscellaneous.inc
```diff
@@ -49,7 +49,20 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
+
+    
+    if ($_bee_options['uri']) {
+      $uri_parts = parse_url($_bee_options['uri']);
+      $_SERVER['HTTP_HOST'] = $uri_parts['host'] . ($uri_parts['port'] ? ':' . $uri_parts['port'] : '');
```
yorkshire-pudding, line 56:
```suggestion
      $_SERVER['HTTP_HOST'] = $uri_parts['host'] . (isset($uri_parts['port']) ? ':' . $uri_parts['port'] : '');
```
This gets rid of the errors if no port is used.

yorkshire-pudding (changes requested):
@quicksketch 
The suggested change fixes the error I encountered at https://github.com/backdrop-contrib/bee/issues/195#issuecomment-1220485663

Please also see my query regarding URI vs URL at the same comment.  I don't mind if there is a good reason for using `uri` such as continuity for `drush` users, but if not, I think `url` may be clearer and would be consistent with the language used in `settings.php`.

yorkshire-pudding:
Agreed with @quicksketch to close this in favour of new PR from me with revisions included.
See https://github.com/backdrop-contrib/bee/pull/276

Issue #189, opened 2022-06-15
bee dl: Uncaught ArgumentCountError function bee_error_handler() when there are typos in a module name


indigoxela:
Ran into this by accident, but it might be worth fixing:

When trying to download a module, but having a typo in the module name, the error handler chokes.

An example:
```
# bee dl foobartypo

There is no release for 'backdrop-contrib/foobartypo'. Do you you want to download the dev version instead? (y/N): y

# y and enter

PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function bee_error_handler(), 4 passed in .../commands/download.bee.inc on line 182 and exactly 5 expected in .../bee.php:44
Stack trace:

#0 ...commands/download.bee.inc(182): bee_error_handler()
#1 ...commands/download.bee.inc(61): download_bee_git_info()
#2 ...includes/command.inc(88): download_bee_callback()
#3 ...bee.php(23): bee_process_command()
#4 {main}
  thrown in ...bee.php on line 44
```
Also with the latest dev (I'm currently on b29e71d5099661e1076cee8290274edfd4d6214e).

Typos can happen easily, making `dl` a little more robust (or maybe the error handler?) might make sense.

yorkshire-pudding:
I've encountered this a few times and thought it could be better.  It should check to see if it can actually find a project first and then gracefully exit with an informative message that the project doesn't exist.

herbdool:
Might be because of: https://www.php.net/manual/en/function.set-error-handler.php:

> Warning
> This parameter has been DEPRECATED as of PHP 7.2.0, and REMOVED as of PHP 8.0.0. If your function defines this parameter without a default, an error of "too few arguments" will be raised when it is called.

I got the fatal error in PHP 8.1 (when testing with `bee cron`). 

This fixes it:

```
function bee_error_handler($errno, $message, $filename = NULL, $line = NULL, $context = NULL) {
  if (error_reporting() > 0) {
    echo "$message\n";
    echo " $filename:$line\n\n";
  }
}
```

ghost:
Thanks @herbdool!

@indigoxela, @yorkshire-pudding: Please see if https://github.com/backdrop-contrib/bee/pull/198 fixes this for you.

indigoxela:
> Please see if https://github.com/backdrop-contrib/bee/pull/198 fixes this for you.

Not yet, but the error changes.

Before:
```
[...]

PR #200, opened 2022-08-12
Make context variable optional by adding a default value

ElusiveMind:
Add in additional error handling to help overly verbose warnings
Update default value to use longform array syntax.

### DIFF
--- a/bee.php
+++ b/bee.php
@@ -41,9 +41,19 @@ exit();
  *
  * @see https://www.php.net/manual/en/function.set-error-handler.php
  */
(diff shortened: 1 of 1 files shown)


yorkshire-pudding:
@ElusiveMind - as per the comment in #189 - I've gone with the simpler approach for now, combined with an improvement to validation.  At the moment I'm not minded to hide errors as its technically still a beta product and we're still trying to eliminate all the bugs.  Hiding them, without them going to a log, might confuse users.  I'll close this PR, but the input is appreciated.

PR #201, opened 2022-08-12
Add a command to drop and recreate the configured database

### LINKED ISSUE #202: Add the Drush db-drop command
I have created Pull request #201 to add the db-drop command to Bee. This also allows Bee to be bootstrapped with an empty database for the purposes of creating tables. I welcome comments and feedback as to the approach.

ElusiveMind:
Fixes #202 
Allow bee to bootstrap with an empty database

### DIFF
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -134,8 +134,8 @@ function db_drop_bee_callback($arguments, $options) {
   if (!$_bee_yes_mode) {
     // Prompt to continue.
     if (!bee_confirm(bt('Are you sure you want to drop the contents of !database ?', array(
-        '!database' => $db_database,
-      )), FALSE)) {
+      '!database' => $db_database,
+    )), FALSE)) {
       return;
     }
   }
@@ -175,7 +175,7 @@ function db_drop_bee_callback($arguments, $options) {
     )), 'success');
   }
   // Remove the temporary mysql options file.
-  bee_delete($connection_file);  
+  bee_delete($connection_file);
 }
 
 /**
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -16,6 +16,16 @@ function db_bee_command() {
         'file' => bt('The SQL file where the exported database will be saved. Leave blank to use the current date/time as the filename.'),
       ),
       'optional_arguments' => array('file'),
+      'options' => array(
+        'column-statistics' => array(
+          'description' => bt("Allow extra options to be passed to the MySQL dump command."),
+          'value' => bt('String'),
+        ),
+        'no-tablespaces' => array(
+          'description' => bt("Allow extra options to be passed to the MySQL dump command."),
+          'value' => bt('String'),
+        ),
+      ),
       'aliases' => array('dbex', 'db-dump', 'sql-export', 'sql-dump'),
       'bootstrap' => BEE_BOOTSTRAP_DATABASE,
       'examples' => array(
@@ -37,12 +47,18 @@ function db_bee_command() {
       ),
     ),
     'db-drop' => array(
-      'description' => bt('Drop the current backdrop database'),
+      'description' => bt('Drop the current database and recreate an empty database with the same details. This could be used prior to import if the target database has more tables than the source database.'),
       'callback' => 'db_drop_bee_callback',
-      'aliases' => array('sql-drop', 'dbd'),
+      'aliases' => array('sql-drop'),
       'bootstrap' => BEE_BOOTSTRAP_DATABASE,
       'examples' => array(
-        'bee sql-drop' => bt('Drop the existing backdrop database. Used when wanting to re-import fresh.'),
+        'bee db-drop' => bt('Drop the current database and recreate an empty database with the same details.'),
+      ),
+      'options' => array(
+        'y' => array(
+          'description' => bt("A database connection string of the form: 'mysql://user:pass@localhost/database_name'. See https://api.backdropcms.org/database-configuration for more details."),
+          'value' => bt('String'),
+        ),
       ),
     ),
     'sql' => array(
@@ -59,48 +75,6 @@ function db_bee_command() {
 }
 
 /**
- * Command callback: Drop and re-create the existing configured database.
- */
-function db_drop_bee_callback($arguments, $options) {
-  // Initialize our pipes variable.
-  $pipes = array();
-
-  // Get database info.
-  $db_connection = Database::getConnectionInfo();
-  $db_info = $db_connection['default'];
-
-  // Drop the existing backdrop database as configured.
-  $command = 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' -e "drop database ' . $db_info['database'] . '"';
-  $result = proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
-
-  if ($result == -1) {
-    bee_message(bt("The '!database' database could not be dropped.", array(
-      '!database' => $db_info['database'],
-    )), 'error');
-  }
-  else {
-    bee_message(bt("The '!database' database was successfully dropped.", array(
-      '!database' => $db_info['database'],
-    )), 'success');
-  }
-
-  // Re-create the existing backdrop database as configured.
-  $command = 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' -e "create database ' . $db_info['database'] . '"';
-  $result = proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
-
-  if ($result == -1) {
-    bee_message(bt("The '!database' database could not be created.", array(
-      '!database' => $db_info['database'],
-    )), 'error');
-  }
-  else {
-    bee_message(bt("The '!database' database was successfully created.", array(
-      '!database' => $db_info['database'],
-    )), 'success');
-  }
-}
-
-/**
  * Command callback: Export the database as a compressed SQL file.
  */
 function db_export_bee_callback($arguments, $options) {
@@ -113,8 +87,14 @@ function db_export_bee_callback($arguments, $options) {
   // Get the filename for the export.
   $filename = isset($arguments['file']) ? $arguments['file'] : date('Ymd_His') . '.sql';
 
+  // Added handling for these two options. Note that trying to mimic "extra-dumps" as per
+  // drush does not work due to the way arguments and options are parsed. (quotes not respected)
+  // These are needed for some configurations of MariaDB/MySQL to work.
+  $colstats = (!empty($options['column-statistics']) || $option['column-statistics'] == '0') ? '--column-statistics=' . $options['column-statistics'] : NULL;
+  $tablespaces = (!empty($options['no-tablespaces'])) ? '--no-tablespaces' : NULL;
+
   // Export and compress the database.
-  $export_command = 'mysqldump --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'] . ' | gzip > ' . $filename . '.gz';
+  $export_command = 'mysqldump --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' ' . $colstats . ' ' . $tablespaces . '  --host=' . $db_info['host'] . ' ' . $db_info['database'] . ' | gzip > ' . $filename . '.gz';
   exec($export_command, $output, $result);
 
   if ($result === 0) {
@@ -142,12 +122,16 @@ function db_import_bee_callback($arguments, $options) {
     $gzip = TRUE;
   }
 
+  $db_username = rawurldecode($db_info['username']);
+  $db_password = rawurldecode($db_info['password']);
+  $db_database = rawurldecode($db_info['database']);
+
   // Import the database.
   $import_command = '';
   if ($gzip) {
     $import_command .= "gunzip -c $filename | ";
   }
-  $import_command .= 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' ' . $db_info['database'];
+  $import_command .= 'mysql --user=' . $db_username . ' --password=' . $db_password . ' --host=' . $db_info['host'] . ' ' . $db_database;
   if (!$gzip) {
     $import_command .= " < $filename";
   }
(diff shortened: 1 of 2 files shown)


commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
         'bee db-import db.sql.gz' => bt('Extract and import db.sql into the current database.'),
       ),
     ),
+    'db-drop' => array(
+      'description' => bt('Drop the current backdrop database'),
+      'callback' => 'db_drop_bee_callback',
+      'aliases' => array('sql-drop', 'dbd'),
```
yorkshire-pudding: For this command, I would leave out the shortened alias of `dbd` as it is more likely to be accidentally executed.
Drush has `sql-drop` and `db-drop` fits into the bee model.  While other db commands in bee have short aliases they are not as potentially destructive without additional text (e.g. db-import while destructive needs a file to import)

commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
         'bee db-import db.sql.gz' => bt('Extract and import db.sql into the current database.'),
       ),
     ),
+    'db-drop' => array(
+      'description' => bt('Drop the current backdrop database'),
```
yorkshire-pudding: Suggested text:
'Drop the current database and recreate an empty database with the same details. This could be used prior to import if the target database has more tables than the source database.'

commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
...
+      'callback' => 'db_drop_bee_callback',
+      'aliases' => array('sql-drop', 'dbd'),
+      'bootstrap' => BEE_BOOTSTRAP_DATABASE,
+      'examples' => array(
+        'bee sql-drop' => bt('Drop the existing backdrop database. Used when wanting to re-import fresh.'),
```
yorkshire-pudding: The standard for examples is to use the main command, so this would be `bee db-drop`
Once the confirmation is in, it would make sense to add an example using the `-y` or `--yes` global option.
It might also be useful to include a multisite example (e.g. `bee --site=multi_one db-drop`)
The example description should just state what the command will do (not the purpose as that goes into the description above) so for the example you've given it would be 'Drop the existing database and recreate with same details.'

commands/db.bee.inc
```diff
@@ -49,6 +58,48 @@ function db_bee_command() {
   );
 }
 
+/**
```
yorkshire-pudding: We should place the command callback function within the code in the order they appear in `db_bee_command` so `db_drop_bee_callback()` would be between `db_import_bee_callback()` and `sql_bee_callback()`

commands/db.bee.inc
```diff
@@ -49,6 +58,48 @@ function db_bee_command() {
...
+function db_drop_bee_callback($arguments, $options) {
+  // Initialize our pipes variable.
+  $pipes = array();
+
+  // Get database info.
+  $db_connection = Database::getConnectionInfo();
+  $db_info = $db_connection['default'];
+
+  // Drop the existing backdrop database as configured.
+  $command = 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' -e "drop database ' . $db_info['database'] . '"';
```
yorkshire-pudding: Need to use `rawurldecode` on `user`, `password` and `database`
See `db_bee_sql_options()` at the bottom; while this helper function won't work for `drop` and `create` as the `database` string is part of the SQL rather than just part of the connection, it should give you an indication of how to use the strings, and it would make sense to replicate the bits other than `database` from that function so it's easier to read once you get the  `rawurldecode`  function in.  That block would only need to be in once to work for both `drop` and `create`.

commands/db.bee.inc
```diff
@@ -49,6 +58,48 @@ function db_bee_command() {
...
+    )), 'error');
+  }
+  else {
+    bee_message(bt("The '!database' database was successfully dropped.", array(
+      '!database' => $db_info['database'],
+    )), 'success');
+  }
+
+  // Re-create the existing backdrop database as configured.
+  $command = 'mysql --user=' . $db_info['username'] . ' --password=' . $db_info['password'] . ' --host=' . $db_info['host'] . ' -e "create database ' . $db_info['database'] . '"';
```
yorkshire-pudding: See comment above

yorkshire-pudding (changes requested):
Thanks @ElusiveMind - I've left some feedback to help get this up to standard.

commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
...
       ),
     ),
+    'db-drop' => array(
+      'description' => bt('Drop the current database and recreate an empty database with the same details. This could be used prior to import if the target database has more tables than the source database.'),
+      'callback' => 'db_drop_bee_callback',
+      'aliases' => array('sql-drop'),
+      'bootstrap' => BEE_BOOTSTRAP_DATABASE,
+      'examples' => array(
+        'bee db-drop' => bt('Drop the current database and recreate an empty database with the same details.'),
+      ),
```
yorkshire-pudding:
```suggestion
      'examples' => array(
        'bee db-drop' => bt('Drop the current database and recreate an empty database with the same details.  You will then be prompted to confirm.'),
        'bee --yes db-drop' => bt('Drop the current database and recreate an empty database with the same details.  You will NOT be prompted to confirm.'),
      ),
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
   }
 }
 
+/**
+ * Command callback: Drop and re-create the existing configured database.
+ */
+function db_drop_bee_callback($arguments, $options) {
+  global $_bee_yes_mode;
+  // Initialize our pipes variable.
+  $pipes = array();
```
yorkshire-pudding: This is not needed and we should keep code as simple as possible.

```suggestion
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+}
+
 /**
  * Command callback: Open an SQL command-line interface using Backdrop's
  * database credentials.
  */
 function sql_bee_callback($arguments, $options) {
+  // Initialize our pipes variable.
+  $pipes = array();
+
```
yorkshire-pudding: This is not needed and we should keep code as simple as possible

```suggestion
```

yorkshire-pudding (changes requested):
@ElusiveMind 
Thank you Michael.  This is starting to look really good.
I've suggested an extra example to make clear that we can bypass the confirmation.

I've added two suggestions to remove the initialisation of the `$pipes` array.  While it is normally best practice to declare arrays, the [PHP documentation page](https://www.php.net/manual/en/function.proc-open.php) does not include this in any examples.  Additionally, it was not used previously in bee and did not cause a problem that I am aware of and wherever possible we want to keep the coding consistent throughout bee, not changing patterns unless there is a very good reason.

Next step is to add a test to `tests/backdrop/DBCommandsTest.php`; I think we want to check that the success message is received and as previously discussed wrap the test for this in an export and import so it is non-destructive.  Probably want to remove the export once successfully imported.

yorkshire-pudding:
@ElusiveMind - once you're back from your travels, we should look at getting this over the line. We're almost there.

commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
         'bee db-import db.sql.gz' => bt('Extract and import db.sql into the current database.'),
       ),
     ),
+    'db-drop' => array(
+      'description' => bt('Drop the current database and recreate an empty database with the same details. This could be used prior to import if the target database has more tables than the source database.'),
+      'callback' => 'db_drop_bee_callback',
+      'aliases' => array('sql-drop'),
+      'bootstrap' => BEE_BOOTSTRAP_DATABASE,
```
yorkshire-pudding: This line can be removed. None of the database functions require bootstrap so I've remove the others in issue #280 
```suggestion
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+ */
+function db_drop_bee_callback($arguments, $options) {
+  global $_bee_yes_mode;
+  // Initialize our pipes variable.
+  $pipes = array();
+
+  // Get database info.
+  $db_connection = Database::getConnectionInfo();
+  $db_info = $db_connection['default'];
+
```
yorkshire-pudding, line 133: We use database name in a number of place so we should add the database name to a variable so we don't need to keep getting it from the array.
```suggestion
  $db_database = rawurldecode($db_info['database']);

```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+  if (!$_bee_yes_mode) {
+    // Prompt to continue.
+    if (!bee_confirm(bt('Are you sure you want to drop the contents of !database ?', array(
+        '!database' => $db_info['database']
```
yorkshire-pudding: We can now use this variable. Also need a comma after to meet coding standards.

```suggestion
        '!database' => $db_database,
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+      )), FALSE)) {
+      return;
+    }
+  }
+
+  $options = '--user=' . rawurldecode($db_info['username']);
+  $options .= " --password='" .  rawurldecode($db_info['password']) . "'";
+  $options .= ' --host=' . $db_info['host'];
+  $db_database = rawurldecode($db_info['database']);
```
yorkshire-pudding: Following changes from #261 and #282 we can now use the helper function.  Also, we have the database name variable above.

```suggestion
  // Get database connection details.
  $connection_details = db_bee_mysql_options($db_info, FALSE);
  $connection_string = $connection_details['connection'];
  $connection_file = $connection_details['filename'];
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+
+  // Drop the existing backdrop database as configured.
+  $command = 'mysql ' . $options . ' -e "drop database ' . $db_database . '";';
```
yorkshire-pudding: We now use the `$connection_string` variable:

```suggestion
  $command = 'mysql ' . $connection_string . ' -e "drop database ' . $db_database . '";';
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+  $result = proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+
+  if ($result == -1) {
+    bee_message(bt("The '!database' database could not be dropped.", array(
+      '!database' => $db_info['database'],
```
yorkshire-pudding: We can use the `$db_database` variable here:

```suggestion
      '!database' => $db_database,
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+    )), 'error');
+  }
+  else {
+    bee_message(bt("The '!database' database was successfully dropped.", array(
+      '!database' => $db_info['database'],
```
yorkshire-pudding: We can use the `$db_database` variable here:

```suggestion
      '!database' => $db_database,
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+    )), 'success');
+  }
+
+  // Re-create the existing backdrop database as configured.
+  $command = 'mysql ' . $options . ' -e "create database ' . $db_database . '";';
```
yorkshire-pudding: We now use the `$connection_string` variable:

```suggestion
  $command = 'mysql ' . $connection_string . ' -e "create database ' . $db_database . '";';
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+  $result = proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
+
+  if ($result == -1) {
+    bee_message(bt("The '!database' database could not be created.", array(
+      '!database' => $db_info['database'],
```
yorkshire-pudding: We can use the `$db_database` variable here:

```suggestion
      '!database' => $db_database,
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+    )), 'error');
+  }
+  else {
+    bee_message(bt("The '!database' database was successfully created.", array(
+      '!database' => $db_info['database'],
```
yorkshire-pudding: We can use the `$db_database` variable here:

```suggestion
      '!database' => $db_database,
```

commands/db.bee.inc
```diff
@@ -110,11 +119,71 @@ function db_import_bee_callback($arguments, $options) {
...
+    )), 'success');
+  }
+}
```
yorkshire-pudding: Like we do in the other functions (from #261), we can delete the temporary mysql options file:

```suggestion
  }
  // Remove the temporary mysql options file.
  bee_delete($connection_file);  
}
```

includes/miscellaneous.inc
```diff
@@ -93,7 +93,7 @@ function bee_initialize_console() {
 function bee_bootstrap($level) {
   global $_bee_backdrop_installed, $_bee_backdrop_root;
 
-  if (!$_bee_backdrop_installed && $level > BEE_BOOTSTRAP_PAGE_CACHE) {
+  if (!$_bee_backdrop_installed && $level > BEE_BOOTSTRAP_DATABASE) {
```
yorkshire-pudding: This is not needed as a result of #280 

```suggestion
  if (!$_bee_backdrop_installed && $level > BEE_BOOTSTRAP_PAGE_CACHE) {
```

commands/db.bee.inc
```diff
@@ -36,6 +36,15 @@ function db_bee_command() {
         'bee db-import db.sql.gz' => bt('Extract and import db.sql into the current database.'),
       ),
     ),
+    'db-drop' => array(
+      'description' => bt('Drop the current database and recreate an empty database with the same details. This could be used prior to import if the target database has more tables than the source database.'),
+      'callback' => 'db_drop_bee_callback',
+      'aliases' => array('sql-drop'),
```
yorkshire-pudding: Following #272 we should add a group here:

```suggestion
      'callback' => 'db_drop_bee_callback',
      'group' => 'database',
      'aliases' => array('sql-drop'),
```

yorkshire-pudding (changes requested):
@ElusiveMind 
Hi Michael

Following a number of changes to the database operations, namely #261 , #280 and #282 I have revisited this and made a number of further suggestions.

It would be great to get this in and we're not far off.  You can add each suggestion to a batch and then apply, or I can do this.

We also need a test for this and I don't think it should be too onerous.  Feel free to have a go adding into `tests/backdrop/DBCommandsTest.php` - I can always offer suggestions if not quite right.

Thanks for your patience and hard work on this.

commands/db.bee.inc
```diff
@@ -110,6 +120,64 @@ function db_import_bee_callback($arguments, $options) {
...
+  // Get database info.
+  $db_connection = Database::getConnectionInfo();
+  $db_info = $db_connection['default'];
+  $db_database = rawurldecode($db_info['database']);
+
+  if (!$_bee_yes_mode) {
+    // Prompt to continue.
+    if (!bee_confirm(bt('Are you sure you want to drop the contents of !database ?', array(
+        '!database' => $db_database,
+      )), FALSE)) {
```
yorkshire-pudding: coding standards
```suggestion
      '!database' => $db_database,
    )), FALSE)) {
```

commands/db.bee.inc
```diff
@@ -110,6 +120,64 @@ function db_import_bee_callback($arguments, $options) {
...
+      '!database' => $db_database,
+    )), 'error');
+  }
+  else {
+    bee_message(bt("The '!database' database was successfully created.", array(
+      '!database' => $db_database,
+    )), 'success');
+  }
+  // Remove the temporary mysql options file.
+  bee_delete($connection_file);
```
yorkshire-pudding: coding standards
```suggestion
  bee_delete($connection_file);
```

yorkshire-pudding (changes requested):

PR #206, opened 2022-08-19
Fixes issue #205 - settings.php was being ignored when bee executes.

### LINKED ISSUE #205: Issue regarding settings.php being ignored
When Bee executes, it loads _settings.php_ at the following line of code:

miscellaneous.inc: Line 57:  `backdrop_settings_initialize();`

After this code executes, `backdrop_bootstrap()` is called which re-executes `backdrop_settings_initialize()` and attempts to re-initializes _settings.php_ but does not succeed because Backdrop loads _settings.php_ using `require_once`.

This resets global `$settings` to an empty array and any code that refers to `$settings` will return empty or `NULL`.

The fix is to change line 57 to invoke `backdrop_bootstrap(BACKDROP_BOOTSTRAP_CONFIGURATION);`

This will employ `backdrop_boostrap()` to only load the configuration phase and prevent it from being re-executed later because backdrop will ignore phases that have previously ran.

rbargerhuff:
Fixes https://github.com/backdrop-contrib/bee/issues/205

### DIFF
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -54,7 +54,7 @@ function bee_initialize_console() {
     // Determine if Backdrop is installed or not.
     require_once $_bee_backdrop_root . '/core/includes/bootstrap.inc';
     if (function_exists('backdrop_bootstrap_is_installed')) {
-      backdrop_settings_initialize();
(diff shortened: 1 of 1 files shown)


yorkshire-pudding:
Closing this as was superseded by #238

PR #209, opened 2022-10-03
Fixes #189 - Error compatibility and download typo handling

### LINKED ISSUE #189: bee dl: Uncaught ArgumentCountError function bee_error_handler() when there are typos in a module name

yorkshire-pudding:
Fixes #189

### DIFF
--- a/commands/download.bee.inc
+++ b/commands/download.bee.inc
@@ -63,7 +63,7 @@ function download_bee_callback($arguments, $options) {
     $url = "https://github.com/$organisation/$project";
     $headers = get_headers($url);
     $response = substr($headers[0], 9, 3);
-    if ($response == 400 || $response == 404) {
+    if ($response >= 400) {
       bee_message(bt("The '!project' project repository could not be found. Please check your spelling and try again.", array(
         '!project' => $project,
       )), 'error');
--- a/commands/download.bee.inc
+++ b/commands/download.bee.inc
@@ -58,55 +58,67 @@ function download_bee_callback($arguments, $options) {
   $progress = (!empty($options['hide-progress'])) ? FALSE : TRUE;
 
   foreach ($arguments['projects'] as $project) {
-    $info = download_bee_git_info($project);
-
-    // Get the project type.
-    if (empty($info['type'])) {
-      bee_message(bt("The 'type' of project '!project' could not be determined.", array(
+    // Check if the project exists by trying to get the repo homepage
+    $organisation = 'backdrop-contrib';
+    $url = "https://github.com/$organisation/$project";
+    $headers = get_headers($url);
+    $response = substr($headers[0], 9, 3);
+    if ($response == 400 || $response == 404) {
+      bee_message(bt("The '!project' project repository could not be found. Please check your spelling and try again.", array(
         '!project' => $project,
       )), 'error');
-      return;
     }
     else {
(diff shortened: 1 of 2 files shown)


commands/download.bee.inc
```diff
@@ -58,55 +58,67 @@ function download_bee_callback($arguments, $options) {
...
-
-    // Get the project type.
-    if (empty($info['type'])) {
-      bee_message(bt("The 'type' of project '!project' could not be determined.", array(
+    // Check if the project exists by trying to get the repo homepage
+    $organisation = 'backdrop-contrib';
+    $url = "https://github.com/$organisation/$project";
+    $headers = get_headers($url);
+    $response = substr($headers[0], 9, 3);
+    if ($response == 400 || $response == 404) {
```
indigoxela: Response 400 stands for "client error", though. Why considering that as "not found"?

yorkshire-pudding: Thanks @indigoxela - this was one area I wasn't clear on how broad the scope should be.  Should I just do 404 or make it more broad?  There is another [part of the code](https://github.com/backdrop-contrib/bee/blob/6626cd5515ac7fc56f86fa945bac3febf6f40360/commands/download.bee.inc#L186) there is does `response < 400`; in this case I didn't want to do that as it would then put the exception right at the bottom of the code.  I could do `response >= 400` to be consistent with the other approach or just `response == 404`; is it better to just focus on 404 or should we allow any error equal to or over 400 to stop it with an error message rather than proceeding.  While I'm only checking backdrop-contrib and so 404 is the most likely, what about if github is down in some way which happens more often that you would expect?  If we keep it broad it will still stop the process going on.  What do you think?

indigoxela: > is it better to just focus on 404 or should we allow any error equal to or over 400 to stop it with an error message rather than proceeding...

Looking at function download_bee_git_info() - I think we should follow the same pattern here. Anything below 400 means, that there's probably something we can handle... 

Anything above 400 means: nothing we can do about it (be it 403, 404, 500 or whatever, nothing will provide a functioning download).

Catching each and every http response code in detail seems overdone to me - either we get something to download or not, this bee action isn't meant to do network analysis. :wink: Only my 2c.

yorkshire-pudding: Thanks - that was the way I was thinking.  I've done it `response >= 400` so the exception is not buried away under everything else.

PR #219, opened 2022-10-18
Issue #84: Status command should display more values

### LINKED ISSUE #84: 'core-status' should display more info
Currently, the `core-status` command displays the following information about a Backdrop site:

```
 Backdrop CMS
 Backdrop root
 Database driver
 Database host
 Database name
 Database username
```

Here's the equivalent info from a Drupal site with Drush:

```
Drupal version
Site URI
Database driver
Database hostname
Database username
Database name
Database
Drupal bootstrap
Drupal user
Default theme
Administration theme
PHP configuration
PHP OS
Drush script
Drush version
Drush temp directory
Drush configuration
Drush alias files
Install profile
Drupal root
Drupal Settings File
Site path
File directory path
Temporary file directory path
```

I think we can add a bit more info to `b`...

indigoxela:
Fixes #84

### DIFF
--- a/commands/status.bee.inc
+++ b/commands/status.bee.inc
@@ -189,6 +189,10 @@ function status_bee_callback($arguments, $options) {
         array('value' => ($config->get('preprocess_js')) ? 'on' : 'off'),
       );
       $rows[] = array(
+        array('value' => bt('Theme debug')),
+        array('value' => ($config->get('theme_debug')) ? 'on' : 'off'),
+      );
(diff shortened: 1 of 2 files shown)


yorkshire-pudding (approved):
Reviewed and tested.  I agree with your rationale for using db_query rather than state_get; if the rest of the function doesn't need the higher bootstrap then it makes sense and it gets the same information.

PR #226, opened 2022-10-24
Issue #225: Catch stat failed for nonexistent update cache dirs

### LINKED ISSUE #225: PHP 8+ warning: stat failed for /tmp/update-cache-XXXX on cron runs
This is a follow-up to #41 and part of the PHP compatibility meta #210.

In php8+ the `@` error operator handling has changed a bit - it will now enforce an error level > 0.

That means that the warnings are back - two for every cron run:

```
stat(): stat failed for /tmp/update-cache-92ff6b81
...core/includes/stream_wrappers.inc:841

stat(): stat failed for /tmp/update-extraction-92ff6b81
...core/includes/stream_wrappers.inc:841
```

See also this related core issue: https://github.com/backdrop/backdrop-issues/issues/5840

I think, we have to catch it in Bee.

indigoxela:
Fixes #225

### DIFF
--- a/bee.php
+++ b/bee.php
@@ -43,6 +43,14 @@ exit();
  */
 function bee_error_handler($errno, $message, $filename, $line, $context = NULL) {
   if (error_reporting() > 0) {
+    // Core uses the @ error operator in url_stat() to suppress the warning for
+    // non-existent files. But that doesn't work for us in PHP 8+.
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Reviewed and tested.

PR #227, opened 2022-10-26
Issue #147: Show Bee version when running the status command

### LINKED ISSUE #147: Show Bee's own version when running the "status" command
For further requests or bug reports here in the queue it would be convenient, if Bee was able to output its own version (like currently 1.x-1.0.0-beta). Drush, for example, shows its version with the `st` command.

Not sure how complex that may be... Just throwing in an idea. :smiley:

indigoxela:
Fixes #147

### DIFF
--- a/commands/status.bee.inc
+++ b/commands/status.bee.inc
@@ -44,6 +44,10 @@ function status_bee_callback($arguments, $options) {
       );
     }
     $rows[] = array(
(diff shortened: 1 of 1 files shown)


yorkshire-pudding (approved):
Tested and reviewed. Thanks @indigoxela

PR #236, opened 2022-11-09
Issue #231: Error display and core function url_stat

### LINKED ISSUE #231: Re-evaluate error_reporting and error display, especially with PHP 8
Since PHP 8 the behavior of the Error Control Operator (`@`) has changed, which is why I see more and more irrelevant errors.

> Prior to PHP 8.0.0, the error_reporting() called inside the custom error handler always returned 0 if the error was suppressed by the @ operator. As of PHP 8.0.0, it returns the value E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE.

Sounds harmless but... All operations that try to flush some stale files or do some cleanup on cron runs, that need to check if such a file exists, now tend to throw errors.

We fixed that partly (for one command) in https://github.com/backdrop-contrib/bee/blob/1.x-1.x/bee.php#L50, but that's not sufficient.
The reason is in core: https://github.com/backdrop/backdrop-issues/issues/5840

So, we know that every `@what_ever_function()` will set the error handler to display at least ... errors - like `stat() failed`.

Currently Bee completely relies on error_reporting() to decide if it should display the message or not. And that will never return `0` for any file operation on non-existent files that ends up in core function url_stat().

Another example that causes nagging: `bee cc all` when the Color module is on. Fixing these problems one by one seems too inefficient.

My suggestion is to re-evaluate how error_reporting is handled in bee_error_handler().

indigoxela:
Fixes #231

### DIFF
--- a/bee.php
+++ b/bee.php
@@ -44,8 +44,9 @@ exit();
 function bee_error_handler($errno, $message, $filename, $line, $context = NULL) {
   if (error_reporting() > 0) {
     // Core uses the @ error operator in url_stat() to suppress the warning for
-    // non-existent files. But that doesn't work for us, as since PHP 8+ the
-    // error_reporting value raises automatically with "@".
+    // non-existent files. However, since PHP 8.0, certain errors are no longer
+    // suppressed which causes unnecessary error messages to appear in some
+    // 'bee' commands.
     if (version_compare(PHP_VERSION, '8', '>=')) {
       $trace = debug_backtrace();
       if (isset($trace[2]) && $trace[2]['function'] == 'url_stat') {
--- a/bee.php
+++ b/bee.php
@@ -44,10 +44,11 @@ exit();
 function bee_error_handler($errno, $message, $filename, $line, $context = NULL) {
   if (error_reporting() > 0) {
     // Core uses the @ error operator in url_stat() to suppress the warning for
-    // non-existent files. But that doesn't work for us in PHP 8+.
+    // non-existent files. But that doesn't work for us, as since PHP 8+ the
+    // error_reporting value raises automatically with "@".
     if (version_compare(PHP_VERSION, '8', '>=')) {
       $trace = debug_backtrace();
-      if (isset($trace[5]) && $trace[5]['function'] == 'installer_clear_update_disk_cache') {
+      if (isset($trace[2]) && $trace[2]['function'] == 'url_stat') {
         return;
       }
     }


bee.php
```diff
@@ -44,10 +44,11 @@
 function bee_error_handler($errno, $message, $filename, $line, $context = NULL) {
   if (error_reporting() > 0) {
     // Core uses the @ error operator in url_stat() to suppress the warning for
-    // non-existent files. But that doesn't work for us in PHP 8+.
+    // non-existent files. But that doesn't work for us, as since PHP 8+ the
```
yorkshire-pudding, line 47: Wasn't sure about this, so checked with my wife, who is a proofreader, and she suggested this was the clearest way.
```suggestion
    // non-existent files. But that doesn't work for us, as, since PHP 8+, the
```

yorkshire-pudding (changes requested):
Minor change to comment requested.  Otherwise, looks great, and works well.

bee.php
indigoxela: I don't know much about English comma rules (heck, never mastered the German ones), but that looks a bit odd:

 `us, as, `

yorkshire-pudding: I'm not sure, that's why I asked my wife.  Trying to find a suitable phrase that works is tricky.  I went to look at what the documentation says, and am confused as to why we're getting this as [`stat()`](https://www.php.net/manual/en/function.stat.php) is supposed to produce `E_WARNING` on failure which should still be suppressed but isn't.  What about instead of `But that doesn't work for us, as since PHP 8+ the error_reporting value raises automatically with "@".` we do:
`However, since PHP 8.0, certain errors are no longer suppressed which causes unnecessary error messages to appear in some 'bee' functions.`

indigoxela: > However, since PHP 8.0, certain errors are no longer suppressed which causes unnecessary error messages to appear in some 'bee' functions.

That sounds a lot better. :+1: One tiny suggestions, maybe `in some Bee commands` - or do you prefer "function"?

yorkshire-pudding: They are referred to as command in the help and wiki.
https://github.com/backdrop-contrib/bee/wiki/Usage

```php
'help' => array(
      'description' => bt("Provide help and examples for 'bee' and its commands."),
```

yorkshire-pudding: So yes - commands works better

bee.php
```diff
@@ -44,10 +44,11 @@
...
+    // error_reporting value raises automatically with "@".
```
yorkshire-pudding:
```suggestion
    // non-existent files. However, since PHP 8.0, certain errors are no longer
    // suppressed which causes unnecessary error messages to appear in some
    // 'bee' commands.
```

yorkshire-pudding (changes requested):
Updated to reflect the suggested comment wording.

yorkshire-pudding (approved):
All looks good. Thanks @indigoxela

PR #238, opened 2022-11-10
Issue #205 - Fix lost settings after backdrop_bootstrap called

### LINKED ISSUE #205: Issue regarding settings.php being ignored

yorkshire-pudding:
Fixes #205

### DIFF
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -104,7 +104,7 @@ function bee_bootstrap($level) {
     $settings_initial = $settings;
     backdrop_bootstrap(BEE_BOOTSTRAP_CONFIGURATION);
     // Restore settings and merge with any newly created settings.
-    $settings = array_unique(array_merge($settings_initial, $settings));
+    $settings = array_merge($settings_initial, $settings);
     backdrop_bootstrap($level);
 
     // Backdrop sets its own error handler, so reset ours here.
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -26,7 +26,7 @@ function bee_initialize_server() {
  * Initialize Bee.
  */
 function bee_initialize_console() {
-  global $_bee_options, $_bee_backdrop_installed, $_bee_backdrop_root, $_bee_backdrop_site, $_bee_multisites, $_bee_yes_mode, $_bee_debug_mode, $_bee_settings;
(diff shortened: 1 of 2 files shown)


includes/miscellaneous.inc
```diff
@@ -99,7 +99,12 @@ function bee_bootstrap($level) {
   }
 
   if (!empty($_bee_backdrop_root) && function_exists('backdrop_bootstrap')) {
+    // Declare Core settings variable so can store settings.
+    global $settings;
+    $settings_initial = $settings;
     backdrop_bootstrap($level);
+    // Restore settings and merge with any newly created settings.
+    $settings = array_unique(array_merge($settings_initial, $settings));
```
quicksketch: I would suggest restoring the settings immediately after the config step, so it is present for all remaining bootstrap steps.
```suggestion
    backdrop_bootstrap(BEE_BOOTSTRAP_CONFIG);
    // Restore settings and merge with any newly created settings.
    $settings = array_unique(array_merge($settings_initial, $settings));
    backdrop_bootstrap($level);
```

Backdrop's bootstrap remembers its current level and only runs the remaining ones if `backdrop_bootstrap()` is called multiple times.

~Separately, we may have the same problem with the `$config` variable, which would be affected by config overrides specified in settings.php. I believe those are also currently lost.~ Nevermind, it's the `$conf` variable (variable overrides) that I was thinking of. Since those only affect `variable_get()` calls I don't think it's a concern.

yorkshire-pudding: Thanks @quicksketch

I didn't understand this at first, but after digging into the `backdrop_bootstrap` code and the functions that are called at each level this makes some sense as only `_backdrop_bootstrap_configuration`  calls `backdrop_settings_initialize()`.  Another learning experience.

Minor change in my commit as constant should be `BEE_BOOTSTRAP_CONFIGURATION`, not `BEE_BOOTSTRAP_CONFIG`.

PR #249, opened 2022-12-05
Issue #246 - Add ability to handle quoted argument values

### LINKED ISSUE #246: Handle arguments provided in quotes as a single argument
This is related to #208 in where the goal is to add --extra=" { mysqldump flags } " support to the database export. This fix is a general fix for the parsing of command line parameters.

By default, parameters are split by -- and then by =. This creates a problem with you have a command such as:

```bee db-export --extra="--column-statistics=0 --no-tablespaces"```

Instead of treating the --extra items as a single parameter, it is split by -- and then by the equals sign to divide it up into multiple parts which are not understood by Bee.

The proposed fix is to modify `bee_parse_input()` to respect quoted arguments and then properly parse any possible equal signs within those commands in a way to be backward compatible to all existing Bee commands.

Referenced in pull request #247

### LINKED ISSUE #248: log command does not accept option with spaces
When trying to view dblog messages that are of a type that includes spaces (e.g. "page not found" or "access denied"), bee will not return any of them. 

The reason is that only the first word (e.g. "page" or "access") are passed as options to the callback. 

To test this out, try this:

```
bee.php --type='page not found'
or
bee.php --type='access denied'
```

yorkshire-pudding:
Fixes #246 
Fixes #248

### DIFF
--- a/includes/command.inc
+++ b/includes/command.inc
@@ -22,8 +22,11 @@ function bee_parse_input() {
       $opt_value = TRUE;
 
       // Get the value, if any.
-      if (strpos($opt_name, '=') !== FALSE) {
-        list($opt_name, $opt_value) = explode('=', $opt_name);
+      if (($pos_equals = strpos($arg, '=')) !== FALSE) {
+        // Extract option name from before '=' and ignoring the preceding '--'.
+        $opt_name = substr($arg, 2 , $pos_equals - 2);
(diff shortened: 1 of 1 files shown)


kbahey (approved):
This patch is verified to fix #248 

Using --type='access denied' or 'page not found' now works as it should.

PR #258, opened 2022-12-15
Issue #191: Add php-script command

### LINKED ISSUE #191: Add support for Drush's `php-script` command
From today's discussion at Backdrop LIVE:

> @jromine: I use "drush php-script foo.php" frequently. Does Bee have that?

AFAICT It looks like this command is not yet supported, so creating an issue to discus it :)

herbdool:
Fixes https://github.com/backdrop-contrib/bee/issues/191


tests/backdrop/ScriptCommandsTest.php
```diff
@@ -0,0 +1,46 @@
...
+    exec('rm /app/backdrop/simple.php');
+
+    // Create a script that calls a Backdrop function.
+    exec('echo "<?php config_set(\"test.settings\", \"value\", \"foobar\");" > /app/backdrop/backdrop-function.php');
+
+    $this->assertTrue(file_exists('/app/backdrop/backdrop-function.php'));
+
+    // Test that calling a Backdrop function works.
+    exec('bee php-script /app/backdrop/backdrop-function.php');
+    $output_backdrop = shell_exec('bee eval \'echo config_get("test.settings", "value")\'');
```
yorkshire-pudding: We should use the bee functions if they are available and don't compromise the test.
```suggestion
    $output_backdrop = shell_exec('bee config-get test.settings value');
```

tests/phpunit.xml
```diff
@@ -10,6 +10,7 @@
       <file>backdrop/DownloadCommandsTest.php</file>
       <file>backdrop/HelpCommandsTest.php</file>
       <file>backdrop/PHPCommandsTest.php</file>
+      <file>backdrop/ScriptCommandsTest.php</file>
```
yorkshire-pudding: The convention is for alphabetical order; I appreciate you put it there as its closely related to the previous command.  If you accept my wider suggestion then won't be needed.

yorkshire-pudding (changes requested):
I've made a minor specific comment around the test.  In general everything looks good and works well.  However, for clarity (as it can't be used for running any other type of script), please can the  command and tests be added to the `php.bee.inc` and `PHPCommandsTest.php` files respectively; we have a precedent for having related functions in the same file.

herbdool:
@yorkshire-pudding I've updated the PR as per your request.

yorkshire-pudding (approved):
Thanks @herbdool - that all looks good.

PR #262, opened 2023-01-11
Add no-tablespaces option to db-export by default

### LINKED ISSUE #260: Include --no-tablespaces option by default in db-export
`--no-tablespaces` is needed in `mysqldump` for versions of MySQL from 5.7.31 and 8.0.21 unless the user (which will be the backdrop database user) has the `PROCESS` privilege, which is not available on cPanel based hosting and possibly others.

Therefore, we should include this by default within the `db-export` command.
It is supported by MariaDB from version 10.2.

We could provide an option to not include this option this if there was a need for it.

(Note, we currently have an option to add this option in via the `extra` option:
```
bee db-export --extra="--no-tablespaces" ../db_export.sql
```

Any thoughts @BWPanda @indigoxela @ElusiveMind @quicksketch @jenlampton @herbdool @klonos 
Tagging a number of people who have engaged with the issue queue lately

yorkshire-pudding:
Fixes #260

### DIFF
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -10,7 +10,7 @@
 function db_bee_command() {
   return array(
     'db-export' => array(
-      'description' => bt('Export the database as a compressed SQL file.'),
+      'description' => bt('Export the database as a compressed SQL file. This uses the --no-tablespaces option by default.'),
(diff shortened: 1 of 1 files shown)


yorkshire-pudding:
Re-opening to re-initiate tests

PR #267, opened 2023-03-04
Issue #266 - Change db-export message to account for absolute and relative paths

### LINKED ISSUE #266: db-export command shows the wrong location if absolute path is used
Upon successful completion of the db-export command, the comment at line 144 prepends the Backdrop Root to the $filename for some reason, so the resulting comment looks as if the saved database is in a folder off the Backdrop Root which it is not.
So I think that line 114 should read as follows:
`'!filename' => $filename,`

yorkshire-pudding:
Fixes #266

### DIFF
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -109,9 +109,15 @@ function db_export_bee_callback($arguments, $options) {
   )), 'log');
 
   if ($result === 0) {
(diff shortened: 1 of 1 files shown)


TheMetMan:
@yorkshire-pudding, I have tested this with and without an absolute path, and the message now tells the truth about the location of the dumped database.
Many thanks for fixing this so quickly.
Regards

PR #275, opened 2023-03-24
Issue #274: Better display for long example commands.

### LINKED ISSUE #274: Allow for long example commands
As a follow-up to https://github.com/backdrop-contrib/bee/issues/161, this issue is about finding a better solution for displaying long example commands.

The current situation is less than ideal:
https://github.com/backdrop-contrib/bee/blob/7099da12268d2455da7033d82ce95f96799ada8e/commands/install.bee.inc#L72-L81
It's hard to read/edit, and (if trying to parse the code) produces weird arrays with part of the command in one key and the other part of the command in another key.

Let's go with @yorkshire-pudding's original suggestion of putting the example command on one line and the example description on the next line.

ghost:
Fixes https://github.com/backdrop-contrib/bee/issues/274

### DIFF
--- a/commands/help.bee.inc
+++ b/commands/help.bee.inc
@@ -171,7 +171,6 @@ function help_bee_command_help(array $descriptor) {
         'value' => $descriptor['command'],
         '#bold' => TRUE,
       ),
-      'newline' => TRUE,
     ),
   );
   if (!empty($aliases)) {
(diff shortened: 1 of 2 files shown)


yorkshire-pudding (approved):
Looks great thanks. A big improvement for the examples in help.

PR #276, opened 2023-03-27
Issue #195: Add base-url option

### LINKED ISSUE #195: Allow setting HTTP_HOST directly via --uri option

yorkshire-pudding:
Fixes #195

### DIFF
--- a/includes/globals.inc
+++ b/includes/globals.inc
@@ -16,6 +16,9 @@ $_bee_global_options = array(
   'site' => array(
     'description' => bt("Specify the directory name or URL of the Backdrop site to use (as defined in 'sites.php'). If not set, will try to find the Backdrop site automatically based on the current directory."),
   ),
+  'base-url' => array(
+    'description' => bt('Specify the base URL of your Backdrop site, such as https://example.com/backdrop.'),
+  ),
   'yes' => array(
     'description' => bt("Answer 'yes' to questions without prompting."),
     'short' => 'y',
--- a/includes/globals.inc
+++ b/includes/globals.inc
@@ -17,7 +17,7 @@ $_bee_global_options = array(
     'description' => bt("Specify the directory name or URL of the Backdrop site to use (as defined in 'sites.php'). If not set, will try to find the Backdrop site automatically based on the current directory."),
   ),
   'base-url' => array(
-    'description' => bt('Specify the base URL of your Backdrop site, such as https://example.com/backdrop.'),
+    'description' => bt("Specify the base URL of the Backdrop site, such as https://example.com/backdrop. May be useful with commands that output URLs to pages on the site."),
   ),
   'yes' => array(
     'description' => bt("Answer 'yes' to questions without prompting."),
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -51,6 +51,9 @@ function bee_initialize_console() {
     define('BACKDROP_ROOT', $_bee_backdrop_root);
 
     if (!empty($_bee_options['base-url'])) {
+      // Strip any trailing slash.
+      $_bee_options['base-url'] = rtrim($_bee_options['base-url'], '/');
+
       $url_parts = parse_url($_bee_options['base-url']);
       if (isset($url_parts['host'])) {
         $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
@@ -58,8 +61,8 @@ function bee_initialize_console() {
       if (isset($url_parts['scheme'])) {
         $_SERVER['HTTPS'] = ($url_parts['scheme'] === 'https');
       }
-      // Set the Backdrop global base_url to ensure that if an https base-url
-      // is specified then an https login link is returned.
+      // Pass the base URL to Backdrop's global variable. This removes the need
+      // to set it manually in settings.php.
       $GLOBALS['base_url'] = $_bee_options['base-url'];
     }
     else {
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -49,7 +49,19 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
+
+    if (isset($_bee_options['base-url'])) {
+      $url_parts = parse_url($_bee_options['base-url']);
+      $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
+      $_SERVER['HTTPS'] = $url_parts['scheme'] === 'https';
+      $GLOBALS['base_url'] = $_bee_options['base-url'];
+    }
+    elseif ($_bee_backdrop_site) {
+      $_SERVER['HTTP_HOST'] = $_bee_backdrop_site;
+    }
+    else {
+      $_SERVER['HTTP_HOST'] = basename($_bee_backdrop_root);
+    }
 
     // Determine if Backdrop is installed or not.
     require_once $_bee_backdrop_root . '/core/includes/bootstrap.inc';
--- a/includes/miscellaneous.inc
+++ b/includes/miscellaneous.inc
@@ -50,19 +50,21 @@ function bee_initialize_console() {
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
 
-    if (isset($_bee_options['base-url'])) {
+    if (!empty($_bee_options['base-url'])) {
       $url_parts = parse_url($_bee_options['base-url']);
-      $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
-      $_SERVER['HTTPS'] = $url_parts['scheme'] === 'https';
+      if (isset($url_parts['host'])) {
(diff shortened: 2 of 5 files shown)


includes/globals.inc
```diff
@@ -16,6 +16,9 @@ $_bee_global_options = array(
   'site' => array(
     'description' => bt("Specify the directory name or URL of the Backdrop site to use (as defined in 'sites.php'). If not set, will try to find the Backdrop site automatically based on the current directory."),
   ),
+  'base-url' => array(
+    'description' => bt('Specify the base URL of your Backdrop site, such as https://example.com/backdrop.'),
```
ghost: Let's keep consistency with the other global options here:
- Use double quotes for the description
- Say 'the site' instead of 'your site'

Let's also explain why this might be useful.

```suggestion
    'description' => bt("Specify the base URL of the Backdrop site, such as https://example.com/backdrop. May be useful with commands that output URLs to pages on the site."),
```

includes/miscellaneous.inc
```diff
@@ -49,7 +49,19 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
+
+    if (isset($_bee_options['base-url'])) {
+      $url_parts = parse_url($_bee_options['base-url']);
+      $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
+      $_SERVER['HTTPS'] = $url_parts['scheme'] === 'https';
+      $GLOBALS['base_url'] = $_bee_options['base-url'];
```
ghost, line 67: What's this used for? Seems unnecessary and confusing...

```suggestion
```

includes/miscellaneous.inc
```diff
@@ -49,7 +49,19 @@ function bee_initialize_console() {
...
+    }
+    elseif ($_bee_backdrop_site) {
+      $_SERVER['HTTP_HOST'] = $_bee_backdrop_site;
+    }
+    else {
+      $_SERVER['HTTP_HOST'] = basename($_bee_backdrop_root);
```
ghost:
```suggestion
    else {
      $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
```

tests/backdrop/UserCommandsTest.php
```diff
@@ -35,9 +35,9 @@ public function test_user_password_command_works() {
    * Make sure that the user-login command works.
    */
   public function test_user_login_command_works() {
-    $output = shell_exec('bee user-login admin');
+    $output = shell_exec('bee --base-url=https://bee.lndo.site user-login admin');
     $this->assertStringContainsString("Use the following link to login as 'admin':", (string) $output);
-    $this->assertStringContainsString('/user/reset/1/', (string) $output);
+    $this->assertStringContainsString('https://bee.lndo.site/user/reset/1/', (string) $output);
```
ghost: I don't think this test should use this global option. It should work without it. If we want to test the global option, we should add a new test for it, ideally in `BeeCoreTest.php` with the tests for the other global options.

ghost (changes requested):

includes/miscellaneous.inc
```diff
@@ -49,7 +49,19 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
+
+    if (isset($_bee_options['base-url'])) {
+      $url_parts = parse_url($_bee_options['base-url']);
+      $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
+      $_SERVER['HTTPS'] = $url_parts['scheme'] === 'https';
```
ghost:
```suggestion
    if (!empty($_bee_options['base-url'])) {
      $url_parts = parse_url($_bee_options['base-url']);
      if (isset($url_parts['host'])) {
        $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
      }
      if (isset($url_parts['scheme'])) {
        $_SERVER['HTTPS'] = ($url_parts['scheme'] === 'https');
      }
```

ghost (changes requested):

includes/miscellaneous.inc
yorkshire-pudding: This sets the Backdrop Global variable and is needed.  Consider the following scenario:
With this line
```shell
$ lando bee uli --base-url=https://bee.lndo.site
Use the following link to login as 'admin':
https://bee.lndo.site/user/reset/1/1680078235/JbNhKhgKWgt7OuLg6l6wlVbrYW-IpiJIYJxjCIxIsUI/login
```

Without this line
``` shell
$ lando bee uli --base-url=https://bee.lndo.site
Use the following link to login as 'admin':
http://bee.lndo.site/user/reset/1/1680078127/o5NfvHUVM2QBJesM61zVxV6NUpq3fGU8i7CKuJQGVFg/login
```

In both tests I put `--base-url=https://bee.lndo.site` but without the line it comes back as `http://bee.lndo.site/user/reset/1/1680078127/o5NfvHUVM2QBJesM61zVxV6NUpq3fGU8i7CKuJQGVFg/login`

Therefore that line is needed for the link to give a valid `https` url.

yorkshire-pudding: I've added a comment to explain that.

includes/globals.inc
yorkshire-pudding: Done

includes/miscellaneous.inc
yorkshire-pudding: Done

tests/backdrop/UserCommandsTest.php
yorkshire-pudding: Done

includes/miscellaneous.inc
yorkshire-pudding: Done

includes/miscellaneous.inc
ghost: Ah, I see. This is how the whole thing works basically. `settings.php` can set this, but this is what Nate was trying to avoid. So setting this global variable is essentially setting the value that `settings.php` defines.

includes/miscellaneous.inc
```diff
@@ -49,8 +49,22 @@ function bee_initialize_console() {
...
+    if (!empty($_bee_options['base-url'])) {
+      $url_parts = parse_url($_bee_options['base-url']);
+      if (isset($url_parts['host'])) {
+        $_SERVER['HTTP_HOST'] = $url_parts['host'] . (isset($url_parts['port']) ? ':' . $url_parts['port'] : '');
+      }
+      if (isset($url_parts['scheme'])) {
+        $_SERVER['HTTPS'] = ($url_parts['scheme'] === 'https');
+      }
+      // Set the Backdrop global base_url to ensure that if an https base-url
+      // is specified then an https login link is returned.
```
ghost: Now that I understand this is what makes the whole thing work, I think we can simplify the comment:

```suggestion

      // Pass the base URL to Backdrop's global variable. This removes the need
      // to set it manually in settings.php.
```

includes/miscellaneous.inc
```diff
@@ -49,8 +49,22 @@ function bee_initialize_console() {
     chdir($_bee_backdrop_root);
     $_bee_backdrop_root = getcwd();
     define('BACKDROP_ROOT', $_bee_backdrop_root);
-    $_SERVER['HTTP_HOST'] = !empty($_bee_backdrop_site) ? $_bee_backdrop_site : basename($_bee_backdrop_root);
 
+    if (!empty($_bee_options['base-url'])) {
+      $url_parts = parse_url($_bee_options['base-url']);
```
ghost: https://github.com/backdrop/backdrop/blob/1.x/settings.php#L147 says there shouldn't be a trailing slash. Indeed, adding one to the Bee command results in a double slash in the URL.

```suggestion
    if (!empty($_bee_options['base-url'])) {
      // Strip any trailing slash.
      $_bee_options['base-url'] = rtrim($_bee_options['base-url'], '/');
      
      $url_parts = parse_url($_bee_options['base-url']);
```

ghost (changes requested):

includes/miscellaneous.inc
yorkshire-pudding: Done

includes/miscellaneous.inc
yorkshire-pudding: Done

ghost (approved):

PR #310, opened 2023-07-13
Issue #289: Replace exec() and shell_exec() in download command

### LINKED ISSUE #289: Provide a user friendly message if shell_exec function is not available
Despite previously enabling it, my host appears to have disabled `shell_exec()` and as the php config was set not to show errors, it was silently failing and once I checked the error log then I discovered what was happening.

Bee should check the presence of this function and gracefully exit with an error message if the command needs this to function.

indigoxela:
Fixes #289 (in an unexpected way :wink:)

### DIFF
--- a/commands/download.bee.inc
+++ b/commands/download.bee.inc
@@ -5,6 +5,11 @@
  */
 
 /**
+ * The useragent string used for HTTP requests via cURL.
+ */
+define('BEE_USERAGENT', 'Bee - the command line tool for Backdrop CMS');
+
+/**
  * Implements hook_bee_command().
  */
 function download_bee_command() {
@@ -296,7 +301,7 @@ function download_bee_git_info($project) {
     if ($answer) {
       $repo_info = FALSE;
       if ($curl_handle = curl_init("$api_url/$organisation/$project")) {
-        curl_setopt($curl_handle, CURLOPT_USERAGENT, 'Bee');// github nag @todo constant
+        curl_setopt($curl_handle, CURLOPT_USERAGENT, BEE_USERAGENT);
         curl_setopt($curl_handle, CURLOPT_HEADER, 0);
         curl_setopt($curl_handle, CURLOPT_RETURNTRANSFER, 1);
         $repo_info = curl_exec($curl_handle);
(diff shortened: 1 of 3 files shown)


commands/download.bee.inc
```diff
@@ -298,9 +294,20 @@ function download_bee_git_info($project) {
     )));
 
     if ($answer) {
+      $repo_info = FALSE;
+      if ($curl_handle = curl_init("$api_url/$organisation/$project")) {
+        curl_setopt($curl_handle, CURLOPT_USERAGENT, 'Bee');// github nag @todo constant
```
yorkshire-pudding: how about 'Bee - the command line tool for Backdrop CMS' ? It works (I've tested that variant) and is clear what is and what it relates to if they ever try to block dodgy user agents, then this should be descriptive enough not to get blocked because they don't know what it is.

Were you thinking about a global constant or a constant at the command level. I'm easy either way.

commands/download.bee.inc
```diff
@@ -19,7 +19,7 @@ function download_bee_command() {
       'multiple_argument' => 'projects',
       'options' => array(
         'hide-progress' => array(
-          'description' => bt('Hide the download progress bar.'),
+          'description' => bt('Deprecated, will get removed in a future version.'),
```
yorkshire-pudding, line 27: We use this in tests MultisiteDownloadCommandsTest and DownloadCammandsTest
While Bee works fine; it just ignores options it doesn't recognise, we should tidy up while we're removing the option.

yorkshire-pudding (changes requested):
Thank you. This works well and I don't think the loss of progress is a problem.  I've left a suggestion for User Agent and also pointed out that we have test functions using the `--hide-progress` option that we can remove this from.

commands/download.bee.inc
indigoxela: > Were you thinking about a global constant or a constant at the command level

Interesting point... I was thinking about a command level constant, AFAIK the download command is the only one, that needs cURL. The rest seems to use core functions.

commands/download.bee.inc
indigoxela: Tests updated. :+1:

To make sure: the deprecation message in descriptions is OK? Or should the option get removed right away?

yorkshire-pudding: Yes, I think the deprecation message makes sense. Bee will pass it to the command callback but it simply won't be used. This will help people know that it is no longer required in their script in the absence of anything like release notes.

yorkshire-pudding (approved):
All looks good. Tests pass locally.
Thank you.

PR #327, opened 2023-09-18
Issue 265: Add db-query command to bee.

### LINKED ISSUE #265: Add the ability to execute SQL queries like drush sql-query command
It would be very useful to be able to execute db queries as in `drush sql-query "SELECT title FROM {node}..."`.

Drush help is:

```
drush help sqlq
Execute a query against a database.

Examples:
 drush sql-query "SELECT * FROM users      Browse user record. Table prefixes, if used, must be added to table names by hand.               
 WHERE uid=1"                                                                                                                               
 drush sql-query --db-prefix "SELECT *     Browse user record. Table prefixes are honored.  Caution: curly-braces will be stripped from all 
 FROM {users} WHERE uid=1"                 portions of the query.                                                                           
 `drush sql-connect` < example.sql         Import sql statements from a file into the current database.                                     
 drush sql-query --file=example.sql        Alternate way to import sql statements from a file.

Arguments:
 query                                     An SQL query. Ignored if 'file' is provided.

Options:
 --database=<key>                          The DB connection key if using multiple connections in settings.php.  
 --db-prefix                               Enable replacement of braces in your query.                           
 --db-url=<mysql://root:pass@127.0.0.1/db  A Drupal 6 style database URL.                                        
 >                                                                                                               
 --extra=<--skip-column-names>             Add custom options to the database connection command.                
 --file                                    Path to a file containing the SQL to be run. Gzip files are accepted. 
 --result-file=</path/to/file>             Save to a file. The file should be relative to Drupal root. Optional.

Aliases: sqlq, sql:query
```

I'll work on this.

bugfolder:
Fixes https://github.com/backdrop-contrib/bee/issues/265.


commands/dbq.bee.inc
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * @file
```
yorkshire-pudding: I think it would make sense to add this to `db.bee.inc`

commands/dbq.bee.inc
```diff
@@ -0,0 +1,59 @@
...
+ * Implements hook_bee_command().
+ */
+function dbq_bee_command() {
+  return array(
+    'db-query' => array(
+      'description' => bt('Execute a query using db_query().'),
+      'callback' => 'dbq_bee_callback',
+      'group' => 'advanced',
+      'arguments' => array(
+        'code' => bt("An SQL query as passed to db_query() (no placeholders)."),
```
yorkshire-pudding:
```suggestion
        'query' => bt("An SQL query as passed to db_query() (no placeholders)."),
```

commands/dbq.bee.inc
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Command callback: Pass to db_query() after bootstrapping Backdrop.
+ *
+ * @see sql.drush.inc
+ */
+function dbq_bee_callback($arguments, $options) {
+  try {
+    $result = db_query($arguments['code'], array());
```
yorkshire-pudding:
```suggestion
    $result = db_query($arguments['query'], array());
```

commands/dbq.bee.inc
```diff
@@ -0,0 +1,59 @@
...
+      foreach ($output as $row) {
+        if (!empty($row)) {
+          $f = fopen('php://memory', 'r+');
+          if (fputcsv($f, $row) === FALSE) {
+              continue;
+          }
+          rewind($f);
+          $csv_line = stream_get_contents($f);
+          $text = rtrim($csv_line);
+          bee_message($text);
```
yorkshire-pudding: I think we would be better off using `bee_display_output()`; this takes the return value from the command and displays as is without a leading icon.  The `state_get_bee_callback()` function has an example of output with a newline. My guess is that this could return multiple lines so each line should be a text value in the array.

yorkshire-pudding (changes requested):
@bugfolder - thanks for this
I think this still fills a useful niche.  I've made a few suggested changes below.  In addition, it would benefit from a test in `tests/backdrop/DBCommandsTest.php`

commands/dbq.bee.inc
bugfolder: OK

commands/dbq.bee.inc
bugfolder: I'll make this change in the next PR.

commands/dbq.bee.inc
bugfolder: I'll make this change in the next PR.

commands/dbq.bee.inc
bugfolder: Will do.

yorkshire-pudding (approved):
Looks good

PR #342, opened 2023-11-19
Issue #341: Keep quieted (@) functions quiet.

### LINKED ISSUE #341: bee doesn't properly handle @some_function() calls
I discovered this when using `bee cc all` on a local site with CiviCRM (cue ominous music) and PHP 8.1.13, which threw this set of errors:

> mysite % bee cc all
>filemtime(): stat failed for /var/www/ousabd/webroot/modules/contrib/civicrm/vendor/composer/installed.json
 /mysite/modules/contrib/civicrm/vendor/symfony/config/Resource/FileResource.php:62
>
>Cannot modify header information - headers already sent by (output started at /myuser/Sites/bee/bee.php:67)
 /mysite/core/includes/bootstrap.inc:1637
>
>Error: Class "Civi\FlexMailer\Services" not found in flexmailer_civicrm_container() (line 50 of /mysite/modules/contrib/civicrm/ext/flexmailer/flexmailer.php).

However, clearing all caches from the UI didn't create the same errors. The first error is thrown from this line of code:

```
        return false !== ($filemtime = @filemtime($this->resource)) && $filemtime <= $timestamp;
```

Note that `@filemtime` is `@`ed, so it shouldn't be throwing an error when the resource is missing.

A simpler way to reproduce the error is to enable the Devel module, then use the "Execute PHP" command to execute this command:

```
echo 'Test: ' . @filemtime('foobar');
```

That will print "Test" with no error, as expected. However, the `bee` equivalent command,

```
bee ev "echo 'Test: ' . @filemtime('foobar');"
```

emits the error.

This is because `bee` has its own error handler, `bee_error_handler()`, which currently does anticipate some problems of this nature:

```
function bee_error_handler($errno, $message, $filename, $line, array $context = NULL) {
  if (error_reporting() > 0) {
    // Core uses the @ error operator in url_stat() to suppress the warning for
    // non-existent files. However, since PHP 8.0, certain errors are no longer
    // suppressed which causes unnecessary error messages to appear in some
    // 'bee' commands.
    if (version_compare(PHP_VERSION, '8', '>=')) {
      $trace = debug_backtrace();
      if (isset($trace[2]) && $trace[2]['function'] == 'url_stat') {
        return;
      }
    }
    echo "$message\n";
    echo " $filename:$line\n\n";
  }
}
```

However, it only considers the core function `url_stat()`. We need it to also consider `filemtime()` (and maybe more?) to work with CiviCRM.

But there's even a problem with this the way it is. Insert `echo var_dump($trace);` just after the `debug_backtrace()`, re-run the problematic `bee` command, and we get this:

```
array(5) {
  [0]=>
  array(2) {
    ["function"]=>
    string(17) "bee_error_handler"
    ["args"]=>
    array(4) {
      [0]=>
      int(2)
      [1]=>
      string(35) "filemtime(): stat failed for foobar"
      [2]=>
      string(64) "/myuser/Sites/bee/commands/php.bee.inc(51) : eval()'d code"
      [3]=>
      int(1)
    }
  }
  [1]=>
  array(4) {
    ["file"]=>
    string(64) "/myuser/Sites/bee/commands/php.bee.inc(51) : eval()'d code"
    ["line"]=>
    int(1)
    ["function"]=>
    string(9) "filemtime"
    ["args"]=>
    array(1) {
      [0]=>
      string(6) "foobar"
    }
  }
  [2]=>
  array(3) {
    ["file"]=>
    string(44) "/myuser/Sites/bee/commands/php.bee.inc"
    ["line"]=>
    int(51)
    ["function"]=>
    string(4) "eval"
  }
  [3]=>
  array(4) {
    ["file"]=>
    string(44) "/myuser/Sites/bee/includes/command.inc"
    ["line"]=>
    int(91)
    ["function"]=>
    string(17) "eval_bee_callback"
    ["args"]=>
    array(2) {
      [0]=>
      array(1) {
        ["code"]=>
        string(37) "echo 'Test: ' . @filemtime('foobar');"
      }
      [1]=>
      array(0) {
      }
    }
  }
  [4]=>
  array(4) {
    ["file"]=>
    string(31) "/myuser/Sites/bee/bee.php"
    ["line"]=>
    int(30)
    ["function"]=>
    string(19) "bee_process_command"
    ["args"]=>
    array(0) {
    }
  }
}
filemtime(): stat failed for foobar
 /myuser/Sites/bee/commands/php.bee.inc(51) : eval()'d code:1
```

Note that the name of the `@`ed function is in `$trace[1]`, not `$trace[2]`. So I don't think this would work even in its original intended use.

So I suggest two changes:

* First, use `$trace[1]`.
* Second, create an array of functions for which this check needs to be made, and use `in_array()` to do the check. Then, as new functions are discovered, they can be added to the array.

A PR will follow shortly.

bugfolder:
Fixes https://github.com/backdrop-contrib/bee/issues/341.

### DIFF
--- a/includes/errors.inc
+++ b/includes/errors.inc
@@ -65,23 +65,24 @@ function _bee_log_error($error, $fatal = FALSE) {
   // backdrop_error_levels. Map this to a bee_message type.
   switch ($error['severity_level']) {
     case WATCHDOG_ERROR:
-      $msg_type = 'error';
+      $type_info = bee_type_info('error');
       break;
 
     case WATCHDOG_WARNING:
-      $msg_type = 'warning';
+      $type_info = bee_type_info('warning');
       break;
 
     case WATCHDOG_NOTICE:
-      $msg_type = 'info';
+      $type_info = bee_type_info('info');
       break;
 
     case WATCHDOG_DEBUG:
-      $msg_type = 'log';
+      $type_info = bee_type_info('log');
       break;
   }
 
-  bee_message(bt("!type: !message\nin !function (line !line of !file).", $error), $msg_type);
+  echo "\n " . bee_format_text($type_info[0], $type_info[1], TRUE) . ' ' . bee_format_text(bt("!type: !message\nin !function (line !line of !file).", $error)) . "\n\n";
+
   if ($fatal) {
     exit();
   }
--- a/includes/errors.inc
+++ b/includes/errors.inc
@@ -51,16 +51,16 @@ function _bee_error_handler_real($error_level, $message, $filename, $line, array
 /**
  * Logs a PHP error or exception to the console.
  *
- * @param $error
+ * @param array $error
  *   An array with the following keys: !type, !message, !function, !file, !line
  *   and severity_level. All the parameters are plain-text.
- * @param $fatal
+ * @param bool $fatal
  *   TRUE if the error is fatal.
  *
  * @see _backdrop_log_error()
  */
 function _bee_log_error($error, $fatal = FALSE) {
-
+x
   // $severity_level is one of the WATCHDOG_* constants defined in
   // backdrop_error_levels. Map this to a bee_message type.
   switch ($error['severity_level']) {
@@ -90,13 +90,13 @@ function _bee_log_error($error, $fatal = FALSE) {
 /**
  * Gets the last caller from a backtrace.
  *
- * @param $backtrace
+ * @param array $backtrace
  *   A standard PHP backtrace.
  *
- * @return
+ * @return array
  *   An associative array with keys 'file', 'line' and 'function'.
  *
- * @see _backdrop_get_last_caller().
+ * @see _backdrop_get_last_caller()
  */
 function _bee_get_last_caller($backtrace) {
   // Errors that occur inside PHP internal functions do not generate
--- a/includes/errors.inc
+++ b/includes/errors.inc
@@ -1,0 +1,126 @@
+<?php
+/**
+ * @file
+ * Error handling for bee.
+ */
+
+/**
+ * Custom error handler for `bee`.
+ *
+ * @param int $error_level
+ *   The level of the error.
+ * @param string $message
+ *   Error message to output to the user.
+ * @param string $filename
+ *   The file that the error came from.
+ * @param int $line
+ *   The line number the error came from.
+ * @param array $context
+ *   An array of all variables from where the error was triggered.
+ *
+ * @see https://www.php.net/manual/en/function.set-error-handler.php
+ * @see _backdrop_error_handler_real()
+ */
+function _bee_error_handler_real($error_level, $message, $filename, $line, array $context = NULL) {
+
+  if ($error_level & error_reporting()) {
+    $types = backdrop_error_levels();
+    list($severity_msg, $severity_level) = $types[$error_level];
+    $caller = _bee_get_last_caller(debug_backtrace());
+
+    if (empty($caller)) {
+      $caller = array(
+        '!function' => bt('<Unknown function>'),
+        '!file' => bt('<Unknown file>'),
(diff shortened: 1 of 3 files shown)


includes/errors.inc
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param array $error
+ *   An array with the following keys: !type, !message, !function, !file, !line
+ *   and severity_level. All the parameters are plain-text.
+ * @param bool $fatal
+ *   TRUE if the error is fatal.
+ *
+ * @see _backdrop_log_error()
+ */
+function _bee_log_error($error, $fatal = FALSE) {
```
yorkshire-pudding: I'm not sure why this isn't being picked up in the workflow,  but it is being picked up in my local. Our phpcs standard requires type hinting where possible and as we have an array type in the doc, we should hint here.

```suggestion
function _bee_log_error(array $error, $fatal = FALSE) {
```

includes/errors.inc
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param array $backtrace
+ *   A standard PHP backtrace.
+ *
+ * @return array
+ *   An associative array with keys 'file', 'line' and 'function'.
+ *
+ * @see _backdrop_get_last_caller()
+ */
+function _bee_get_last_caller($backtrace) {
```
yorkshire-pudding: same as above

```suggestion
function _bee_get_last_caller(array $backtrace) {
```

includes/errors.inc
```diff
@@ -0,0 +1,127 @@
...
+ * @param array $context
+ *   An array of all variables from where the error was triggered.
+ *
+ * @see https://www.php.net/manual/en/function.set-error-handler.php
+ * @see _backdrop_error_handler_real()
+ */
+function _bee_error_handler_real($error_level, $message, $filename, $line, array $context = NULL) {
+
+  if ($error_level & error_reporting()) {
+    $types = backdrop_error_levels();
```
yorkshire-pudding: I think we'll need to replicate `backdrop_error_levels()` in bee (e.g. `bee_error_levels()`) as we can run bee without backdrop being installed, for example, to download core and then install.

I got this locally, presumably triggered by a connection error downloading core:

```
Download Commands
 ✘ Download core command works
   ┐
   ├ Failed asserting that '\n
   ├ Fatal error: Uncaught Error: Call to undefined function backdrop_error_levels() in /app/includes/errors.inc:27\n
   ├ Stack trace:\n
   ├ #0 /app/bee.php(54): _bee_error_handler_real(2, 'get_headers(): ...', '/app/commands/d...', 337, NULL)\n
   ├ #1 [internal function]: bee_error_handler(2, 'get_headers(): ...', '/app/commands/d...', 337)\n
   ├ #2 /app/commands/download.bee.inc(337): get_headers('https://github....')\n
   ├ #3 /app/commands/download.bee.inc(288): download_bee_git_info('backdrop', '')\n
   ├ #4 /app/includes/command.inc(91): download_core_bee_callback(Array, Array)\n
   ├ #5 /app/bee.php(30): bee_process_command()\n
   ├ #6 {main}\n
   ├   thrown in /app/includes/errors.inc on line 27\n
   ├ ' contains "Backdrop was downloaded into '/app/directory'.".
   │
   ╵ /app/tests/backdrop/DownloadCommandsTest.php:46
   ┴

FAILURES!
Tests: 43, Assertions: 118, Failures: 1.
```

yorkshire-pudding (changes requested):
@bugfolder - Sorry this has taken a while to get to

This looks really good. Thank you!  A few minor(ish) points to address

includes/errors.inc
bugfolder: Yes, added. Also added `bee` versions of `WATCHDOG_*` constants, since those, too, would presumably not be available.

PR #349, opened 2023-12-01
Add support for downloading only security updates to the `update` command

### LINKED ISSUE #300: Update Command - Allow command to only download security updates
This is a subtask of https://github.com/backdrop-contrib/bee/issues/111 and relates to the [issue-111-update-command](https://github.com/backdrop-contrib/bee/tree/issue-111-update-command) branch.

As a site developer
I want to run just security updates
So that I can keep my site secure but ensure other updates get properly tested.

anemirovsky:
Fixes #300

Add support for downloading only security updates to the `update` command.

This adds support for a `--security-only` flag to allow updates to be run just for modules with security releases. ~~This also adds a way to report on the avaiable updates without actually running them, as well as an alternate output format that supports just listing the modules names. This mimics the behavior of `drush pm-updatestatus --format=list` which is useful for feeding the output to another process that actually runs the output, like an Ansible playbook.~~

~~I'm not sure if it makes sense to combine the `--report-only` and `--output-format` into the `update` command rather than adding a separate `updatelist` command. I started with a separate one but realized so much of the code is the same. Happy to split it out if that's preferred, though.~~

### DIFF
--- a/commands/update.bee.inc
+++ b/commands/update.bee.inc
@@ -32,10 +32,12 @@ function update_bee_command() {
       'options' => array(
         'security-only' => array(
           'description' => bt('Only update modules that have security updates available.'),
+          'short' => 's',
         ),
       ),
       'examples' => array(
         'bee update' => bt('Update everything with a new release.'),
+        'bee update --security-only' => bt('Update everything that is a security update only'),
         'bee update webform tatsu' => bt('Updates the Webform module and Tatsu theme only'),
       ),
     ),
@@ -154,7 +156,7 @@ function update_bee_callback($arguments, $options) {
         unset($data[$item['name']]);
         continue;
       }
-      if (isset($options['security-only']) && $item['status'] !== UPDATE_NOT_SECURE) {
+      if ((isset($options['security-only']) || isset($options['s'])) && $item['status'] !== UPDATE_NOT_SECURE) {
         unset($data[$item['name']]);
       }
     }
--- a/commands/update.bee.inc
+++ b/commands/update.bee.inc
@@ -33,13 +33,6 @@ function update_bee_command() {
         'security-only' => array(
           'description' => bt('Only update modules that have security updates available.'),
         ),
-        'report-only' => array(
-          'description' => bt('Output a list of modules with available updates rather than updating them.'),
-        ),
-        'output-format' => array(
-          'description' => bt('Format to output update information in. Options are "table" (default), "list".'),
-          'value' => bt('Format'),
-        ),
       ),
       'examples' => array(
         'bee update' => bt('Update everything with a new release.'),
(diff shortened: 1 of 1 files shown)


yorkshire-pudding:
Hi @anemirovsky - thanks for this contribution.  

I like the idea of the update status and this is better than currently where it shows updates and then asks you to confirm, so you enter 'n'.  However, I think it should be a separate command within this update group rather than options in the update command. Please can you remove from this pull request and do a separate pull request for the update status command against issue #350 - an issue I've created for this.

I've had a quick look at the security element and it looks ok, but I'll test and do a thorough review once the other element is removed.

By the way, if you add `Fixes #300` at the top of the body of your first PR comment, it will automatically link to the issue and helps to avoid effort duplication as well as making it easier to find.

anemirovsky:
I've updated this PR to include just the `--security-only` option and will submit a separate PR for #350 soon. Thank you!

commands/update.bee.inc
```diff
@@ -29,6 +29,11 @@ function update_bee_command() {
...
       'optional_arguments' => array('projects'),
       'multiple_argument' => 'projects',
+      'options' => array(
+        'security-only' => array(
+          'description' => bt('Only update modules that have security updates available.'),
+        ),
+      ),
       'examples' => array(
         'bee update' => bt('Update everything with a new release.'),
         'bee update webform tatsu' => bt('Updates the Webform module and Tatsu theme only'),
```
yorkshire-pudding: We should include an example of using this option.

```suggestion
        'bee update' => bt('Update everything with a new release.'),
        'bee update --security-only => bt('Update everything that is a security update only'),
        'bee update webform tatsu' => bt('Updates the Webform module and Tatsu theme only'),
```

commands/update.bee.inc
```diff
@@ -29,6 +29,11 @@ function update_bee_command() {
       ),
       'optional_arguments' => array('projects'),
       'multiple_argument' => 'projects',
+      'options' => array(
+        'security-only' => array(
+          'description' => bt('Only update modules that have security updates available.'),
+        ),
```
yorkshire-pudding: If a command option does not require a value we can include a short version that are used like `-s`.  :

```suggestion
        'security-only' => array(
          'description' => bt('Only update modules that have security updates available.'),
          'short' => 's',
        ),
```

With this setting, it doesn't have to be a single letter. I did wonder whether `-sec` would be better but stuck with single letter. Feel free to change if you prefer.

yorkshire-pudding (changes requested):
I've tested this and it works perfectly. Thank you.
There are a couple of minor enhancements suggested: an example of using this, and adding a short version of this option

Also, I've fixed a bug that arose from a recent change in Backdrop that was causing the tests to fail (minor change in commands/dblog.bee.inc) - if you can rebase your branch so it includes that change then the tests should pass.  At present, we don't have any tests for this feature, but we want to ensure that any changes we make don't break existing functionality.

anemirovsky:
Done, thank you!

PR #367, opened 2024-04-30
Issue #197: Create box.json

### LINKED ISSUE #197: Compile Releases into Phar files
PHP Archives (Phar files) can be very useful to bundle together all the code of a PHP project into a single file. This can make deployment to servers or local environments easier because only a single file needs to be downloaded instead of an archive that expands to a directory. It's also potentially useful because the phar file can act as an executable directly, so there's no need to set up a path alias routing `bee.php` to just `bee`.

It would be nice is `bee` started creating releases on GitHub, and with each release it packaged a `phar` file for download. There are existing Marketplace actions like https://github.com/marketplace/actions/box-packager-for-phar that could be used, or a custom workflow could be created such as the example at https://github.com/sergeyklay/gh-actions-php-phar-create-release-example/tree/master/.github/workflows (both use the same underlying "box" project to create the phar file).

herbdool:
Fixes #197

### DIFF
--- a/box.json
+++ b/box.json
@@ -1,0 +1,14 @@
+{
+  "chmod": "0755",
+  "directories": [
+    "commands",
+    "includes"
+  ],
+  "compactors": [
+    "KevinGH\\Box\\Compactor\\Php"
+  ],
+  "git-version": "package_version",
+  "main": "bee.php",
+  "output": "bee.phar",
+  "stub": true
+}
--- a/box.json
+++ b/box.json
@@ -1,5 +1,4 @@
 {
-  "chmod": "0755",
   "directories": [
     "commands",
     "includes"
@@ -9,6 +8,4 @@
   ],
   "git-version": "package_version",
   "main": "bee.php",
-  "output": "bee.phar",
-  "stub": true
 }


box.json
```diff
@@ -0,0 +1,14 @@
+{
+  "chmod": "0755",
```
yorkshire-pudding, line 2:
```suggestion
```

box.json
```diff
@@ -0,0 +1,14 @@
...
+    "commands",
+    "includes"
+  ],
+  "compactors": [
+    "KevinGH\\Box\\Compactor\\Php"
+  ],
+  "git-version": "package_version",
+  "main": "bee.php",
+  "output": "bee.phar",
+  "stub": true
```
yorkshire-pudding:
```suggestion
```

box.json
yorkshire-pudding:
```suggestion
```

box.json
```diff
@@ -0,0 +1,14 @@
...
+  "directories": [
+    "commands",
+    "includes"
+  ],
+  "compactors": [
+    "KevinGH\\Box\\Compactor\\Php"
+  ],
+  "git-version": "package_version",
+  "main": "bee.php",
+  "output": "bee.phar",
```
yorkshire-pudding:
```suggestion
```

yorkshire-pudding (commented):
Hi @herbdool 

When I run it, I get this

```
💡  3 recommendations found:
    - The "output" setting can be omitted since is set to its default value
    - The "stub" setting can be omitted since is set to its default value
    - The "chmod" setting can be omitted since is set to its default value
```

I'm ambivalent about the output one as it could be helpful to have that be explicit for someone looking at it, but there's probably little benefit in including the other default values.

herbdool:
@yorkshire-pudding sure, let's make all those changes. I've updated it.

yorkshire-pudding (approved):

PR #381, opened 2024-05-20
Issue #380: Support db config storage in status command

### LINKED ISSUE #380: Provide info about config storage backend in status command
Currently, if the active config storage class is db (`$settings['config_active_class'] = 'ConfigDatabaseStorage';`) and the path isn't set, there's no display of that info when running `bee st`.

It would be nice to add that to status.bee.inc

Note: there's absolutely no error or problem, just no info output re db. So it's not urgent, just nice to have.

yorkshire-pudding:
Fixes #380


indigoxela (approved):
LGTM :+1:

PR #387, opened 2024-06-20
Explicitly call the PHP binary to avoid errors when installing the site

### LINKED ISSUE #386: Improve bee site-install to avoid permission error on install.sh script, which often needs execute permissions adding
When attempting to do the site initialization with Bee, it assumes file permissions that don't match that of a just-downloaded Backdrop install:

    /tmp $ mkdir bkdrp
    /tmp $ cd bkdrp/
    /tmp/bkdrp $ bee dl-core
    
     [✔] Backdrop was downloaded into '/tmp/bkdrp'.
    
    /tmp/bkdrp $ bee site-install --auto --db-name foo --db-user foo --db-pass foo
    
    Installing Backdrop. This may take a moment...
    
    sh: 1: ./core/scripts/install.sh: Permission denied

This is easily fixed by issuing a `chmod 0755 core/scripts/install.sh`. I guess a cleaner way would be to modify `install.bee.inc` to explicitly call the PHP binary that's already being used instead of calling the PHP interpreter with the `install.sh` script as an argument.

gwolf:
Fixes #386

### DIFF
--- a/commands/install.bee.inc
+++ b/commands/install.bee.inc
@@ -223,9 +223,8 @@ function install_bee_callback($arguments, $options) {
 
   // Run the install script.
   bee_render_text(array('value' => bt("\nInstalling Backdrop. This may take a moment...\n")));
-  // The core/scripts/install.sh binary is not installed as executable
-  // by default. Call it explicitly via the currently-specified PHP
-  // binary.
+  // The core/scripts/install.sh binary is not installed as executable by
+  // default. Call it explicitly via the currently-specified PHP binary.
   $php_bin = PHP_BINARY;
   $result = exec("cd $_bee_backdrop_root && $php_bin ./core/scripts/install.sh $url $db $db_prefix $username $password $email $site_name $site_mail $langcode $clean_urls $profile");
   // Strip tags to prevent HTML code being output.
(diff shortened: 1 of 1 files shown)


commands/install.bee.inc
```diff
@@ -223,7 +223,11 @@ function install_bee_callback($arguments, $options) {
 
   // Run the install script.
   bee_render_text(array('value' => bt("\nInstalling Backdrop. This may take a moment...\n")));
-  $result = exec("cd $_bee_backdrop_root && ./core/scripts/install.sh $url $db $db_prefix $username $password $email $site_name $site_mail $langcode $clean_urls $profile");
+  // The core/scripts/install.sh binary is not installed as executable
+  // by default. Call it explicitly via the currently-specified PHP
+  // binary.
```
yorkshire-pudding: A minor tweak. Coding standards allow up to 80 characters per line (https://docs.backdropcms.org/php-standards#line-length, https://docs.backdropcms.org/doc-standards#general) and this would benefit from that. Takes down to 2 lines.

```suggestion
  // The core/scripts/install.sh binary is not installed as executable by
  // default. Call it explicitly via the currently-specified PHP binary.
```

yorkshire-pudding (changes requested):
This is much simpler that what I was thinking, but it works.
Minor amendment required for line length of comments.

PR #389, opened 2024-06-20
Check for the presence of the MySQL binaries before calling them

### LINKED ISSUE #388: Does not verify whether the MySQL client is installed
Bee has several commands that require having the MySQL client, which is not a system requirement for Backdrop. It can lead to frustrating errors, such as:

```
/var/www/bkdrp_proglo $ bee sqlc
sh: 1: mysql: not found
✓ gwolf@backdrop『33』/var/www/bkdrp_proglo $ bee db-drop
Are you sure you want to drop the contents of bkdrp_proglo ? (y/N): y
sh: 1: mysql: not found
sh: 1: mysql: not found

 [✔] The 'bkdrp_proglo' database was successfully dropped. 
 [✔] The 'bkdrp_proglo' database was successfully created. 
```

I will be submitting  a pull request proposing a patch that fixes this by checking before exec() / proc_open().

gwolf:
Fixes #388 
Patch to verify for the presence of `mysql` and `mysqldump` (and `gunzip`, as it was in the same command construction) before actually calling it.

Thanks!

### DIFF
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -121,8 +121,7 @@ function db_export_bee_callback($arguments, $options) {
 
   $export_command = executable_for('mysqldump');
   if (!$export_command) {
-    bee_message("The MySQL export command 'mysqldump' cannot be found in " .
-                "this system. Please install it and try again.", 'error');
+      bee_message(bt("The MySQL export command 'mysqldump' cannot be found in this system. Please install it and try again."), 'error');
     return FALSE;
   }
   // Export and compress the database.
@@ -180,16 +179,14 @@ function db_import_bee_callback($arguments, $options) {
   // Import the database.
   $mysql_bin = executable_for('mysql');
   if (!$mysql_bin) {
-    bee_message("The MySQL client command 'mysql' cannot be found in this " .
-                "system. Please install it and try again.", 'error');
+    bee_message(bt("The MySQL client command 'mysql' cannot be found in this system. Please install it and try again."), 'error');
     return FALSE;
   }
   $import_command = '';
   if ($gzip) {
     $gunzip_bin = executable_for('gunzip');
     if (!$gunzip_bin) {
-      bee_message("The gzip decompressor command 'gunzip' cannot be found " .
-                  "in this system. Please install it and try again.", 'error');
+      bee_message(bt("The gzip decompressor command 'gunzip' cannot be found in this system. Please install it and try again."), 'error');
       return FALSE;
     }
     $import_command .= "$gunzip_bin -c $filename | ";
@@ -238,8 +235,7 @@ function db_drop_bee_callback($arguments, $options) {
   // Drop the existing backdrop database as configured.
   $mysql_command = executable_for('mysql');
   if (!$mysql_command) {
-    bee_message("The MySQL client command 'mysql' cannot be found in this " .
-                "system. Please install it and try again.", 'error');
+    bee_message(bt("The MySQL client command 'mysql' cannot be found in this system. Please install it and try again."), 'error');
     return FALSE;
   }
   $command = $mysql_command . ' ' . $connection_string . ' -e "drop database ' . $db_database . '";';
@@ -291,8 +287,7 @@ function sql_bee_callback($arguments, $options) {
   // Open SQL command-line.
   $mysql_command = executable_for('mysql');
   if (!$mysql_command) {
-    bee_message("The MySQL client command 'mysql' cannot be found in this " .
-                "system. Please install it and try again.", 'error');
+    bee_message(bt("The MySQL client command 'mysql' cannot be found in this system. Please install it and try again."), 'error');
     return FALSE;
   }
   $command = $mysql_command . ' ' . $connection_string;
--- a/commands/db.bee.inc
+++ b/commands/db.bee.inc
@@ -119,8 +119,14 @@ function db_export_bee_callback($arguments, $options) {
     $extra = '--no-tablespaces ';
   }
 
+  $export_command = executable_for('mysqldump');
+  if (is_null($export_command)) {
+    bee_message("The MySQL export command 'mysqldump' cannot be found in " .
+                "this system. Please install it and try again.", 'error');
+    return null;
+  }
   // Export and compress the database.
-  $export_command = 'mysqldump ';
+  $export_command .= ' ';
   $export_command .= $connection_string;
   $export_command .= ' ';
   $export_command .= $extra;
@@ -172,11 +178,23 @@ function db_import_bee_callback($arguments, $options) {
   }
 
   // Import the database.
+  $mysql_bin = executable_for('mysql');
+  if (is_null($mysql_bin)) {
+    bee_message("The MySQL client command 'mysql' cannot be found in " .
+                "this system. Please install it and try again.", 'error');
+    return null;
+  }
   $import_command = '';
   if ($gzip) {
-    $import_command .= "gunzip -c $filename | ";
+    $gunzip_bin = executable_for('gunzip');
+    if (is_null($gunzip_bin)) {
+      bee_message("The gzip decompressor command 'gunzip' cannot be found in " .
+                  "this system. Please install it and try again.", 'error');
+      return null;
+    }
+    $import_command .= "$gunzip_bin -c $filename | ";
   }
-  $import_command .= 'mysql ' . $connection_string;
+  $import_command .= $mysql_bin . ' ' . $connection_string;
   if (!$gzip) {
     $import_command .= " < $filename";
   }
@@ -218,7 +236,13 @@ function db_drop_bee_callback($arguments, $options) {
   $connection_file = $connection_details['filename'];
 
   // Drop the existing backdrop database as configured.
-  $command = 'mysql ' . $connection_string . ' -e "drop database ' . $db_database . '";';
+  $mysql_command = executable_for('mysql');
+  if (is_null($mysql_command)) {
+    bee_message("The MySQL client command 'mysql' cannot be found in " .
+                "this system. Please install it and try again.", 'error');
+    return null;
+  }
+  $command = $mysql_command . ' ' . $connection_string . ' -e "drop database ' . $db_database . '";';
   $result = proc_close(proc_open($command, array(STDIN, STDOUT, STDERR), $pipes));
 
   if ($result == -1) {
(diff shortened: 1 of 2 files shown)


commands/db.bee.inc
```diff
@@ -348,3 +378,17 @@ function dbq_bee_callback($arguments, $options) {
     bee_message($e->getMessage(), 'error');
   }
 }
+
+/**
+ * Checks whether a command exists in the PATH, and if so, returns the full
+ * path. Returns null if no such command is available.
+ */
```
yorkshire-pudding: Please follow the guidance for https://docs.backdropcms.org/doc-standards#functions
adding `@param` and `@return`

commands/db.bee.inc
```diff
@@ -348,3 +378,17 @@ function dbq_bee_callback($arguments, $options) {
...
+function executable_for($cmd) {
+  foreach (explode(":", getenv('PATH')) as $dir) {
+    $candidate = "$dir/$cmd";
+    if (is_executable($candidate)) {
+      return $candidate;
+    }
+  }
+  return null;
```
yorkshire-pudding: Please note
https://docs.backdropcms.org/php-standards#naming-conventions

> Constants should always be all-uppercase, with underscores to separate words. This includes pre-defined PHP constants like TRUE, FALSE, and NULL

```suggestion
  return NULL;
```

However, we don't normally use `NULL` as a return value. `FALSE` is simpler as the check for it in this case would be something like:
```php
if (!$cmd_bin) {
```

```suggestion
  return FALSE;
```

commands/db.bee.inc
```diff
@@ -119,8 +119,14 @@ function db_export_bee_callback($arguments, $options) {
     $extra = '--no-tablespaces ';
   }
 
+  $export_command = executable_for('mysqldump');
+  if (is_null($export_command)) {
+    bee_message("The MySQL export command 'mysqldump' cannot be found in " .
+                "this system. Please install it and try again.", 'error');
```
yorkshire-pudding: Within Backdrop, there is a function called `t()` that encloses translatable strings.  In bee, because we can't always rely on Backdrop being bootstrapped, we have our own function called `bt()` for translatable strings. Translatable strings are an exception (there are a couple of others) to the 80 character rule and the translatable string should be in a single string.

```suggestion
    bee_message(bt("The MySQL export command 'mysqldump' cannot be found in this system. Please install it and try again."), 'error');
```

commands/db.bee.inc
```diff
@@ -119,8 +119,14 @@ function db_export_bee_callback($arguments, $options) {
...
+    return null;
```
yorkshire-pudding:
```suggestion
    return FALSE;
```

commands/db.bee.inc
```diff
@@ -172,11 +178,23 @@ function db_import_bee_callback($arguments, $options) {
   }
 
   // Import the database.
+  $mysql_bin = executable_for('mysql');
+  if (is_null($mysql_bin)) {
+    bee_message("The MySQL client command 'mysql' cannot be found in this " .
+                "system. Please install it and try again.", 'error');
```
yorkshire-pudding: Above message about `bt()` applies

commands/db.bee.inc
```diff
@@ -172,11 +178,23 @@ function db_import_bee_callback($arguments, $options) {
...
+    return null;
```
yorkshire-pudding:
```suggestion
    return FALSE;
```

commands/db.bee.inc
```diff
@@ -172,11 +178,23 @@ function db_import_bee_callback($arguments, $options) {
...
+  }
   $import_command = '';
   if ($gzip) {
-    $import_command .= "gunzip -c $filename | ";
+    $gunzip_bin = executable_for('gunzip');
+    if (is_null($gunzip_bin)) {
+      bee_message("The gzip decompressor command 'gunzip' cannot be found " .
+                  "in this system. Please install it and try again.", 'error');
```
yorkshire-pudding: Above message about `bt()` applies

commands/db.bee.inc
```diff
@@ -172,11 +178,23 @@ function db_import_bee_callback($arguments, $options) {
...
+      return null;
```
yorkshire-pudding:
```suggestion
      return FALSE;
```

commands/db.bee.inc
```diff
@@ -218,7 +236,13 @@ function db_drop_bee_callback($arguments, $options) {
   $connection_file = $connection_details['filename'];
 
   // Drop the existing backdrop database as configured.
-  $command = 'mysql ' . $connection_string . ' -e "drop database ' . $db_database . '";';
+  $mysql_command = executable_for('mysql');
+  if (is_null($mysql_command)) {
+    bee_message("The MySQL client command 'mysql' cannot be found in this " .
+                "system. Please install it and try again.", 'error');
```
yorkshire-pudding: Above message about `bt()` applies

commands/db.bee.inc
```diff
@@ -218,7 +236,13 @@ function db_drop_bee_callback($arguments, $options) {
...
+    return null;
```
yorkshire-pudding:
```suggestion
    return FALSE;
```

commands/db.bee.inc
```diff
@@ -265,7 +289,13 @@ function sql_bee_callback($arguments, $options) {
   $connection_file = $connection_details['filename'];
 
   // Open SQL command-line.
-  $command = 'mysql ' . $connection_string;
+  $mysql_command = executable_for('mysql');
+  if (is_null($mysql_command)) {
+    bee_message("The MySQL client command 'mysql' cannot be found in this " .
+                "system. Please install it and try again.", 'error');
```
yorkshire-pudding: Above message about `bt()` applies

commands/db.bee.inc
```diff
@@ -265,7 +289,13 @@ function sql_bee_callback($arguments, $options) {
...
+    return null;
```
yorkshire-pudding:
```suggestion
    return FALSE;
```

yorkshire-pudding (changes requested):
Hi @gwolf 

I haven't yet tested this (I don't have any environments without these on to test).  I will run the automated tests again once the changes are made, but might also see if anyone else is able to test the negative conditions.  Most changes are about using `FALSE` instead of `null`, and making the `bee_message` strings translatable. Also a function docblock required.

Thank you

gwolf:
Thanks!
I think I have addressed the needed changes for the coding standards. Sorry, I'm not used to PHP development, and of course, am completely new to this project.

gwolf:
Ugh, the codesniffer still does not like my changes. And the details it gives are not really useful (just says "Error:" 😕)

gwolf:
All tests passed! \o/

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
     bee_message($e->getMessage(), 'error');
   }
 }
+
+/**
+ * Checks whether a command exists in the PATH, and if so, returns the full
+ * path. Returns FALSE if no such command is available.
```
yorkshire-pudding: In line with my suggestion further down about moving this to `includes/filesystem.inc`, it needs a wider explanation of purpose to help other contributors or people writing their own commands.

```suggestion
 * Get the full path for a command executable if it exists in PATH.
 *
 * If your function calls an executable on the system rather than using `bee`,
 * `backdrop` or `php` functions, you can use this function to get the full
 * path of the executable or return FALSE if the executable doesn't exist. The
 * result can then be checked in your calling function, and you can exit with a
 * user friendly error message if the executable doesn't exist in the system.
```

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
...
+ *
+ * @param string $cmd
+ *   Name of the command to be searched.
+ *
+ * @return string
```
yorkshire-pudding:
```suggestion
 * @return string|FALSE
```

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
...
+ *   Full path to the first occurence of $cmd in the search path.
+ *   Returns FALSE if no matching command is found.
+ */
+function executable_for($cmd) {
```
yorkshire-pudding: I've been giving the name of this function a bit of thought, and I've come to the conclusion that this function actually belongs in `includes/filesystem.inc` (`includes` files are all included in `bee.php`) rather than in the `db.bee.inc` file.  Although only db functions currently use this in the core bee software, people writing their own commands or adding a bee command for their module might be grateful for it.  Please could you name the function `bee_get_executable_path` and move to  `includes/filesystem.inc`  - You'll need to update the calls to this function.

Also, can we have full names for variables please.

```suggestion
function bee_get_executable_path($command) {
```

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
...
+  foreach (explode(":", getenv('PATH')) as $dir) {
```
yorkshire-pudding:
```suggestion
  foreach (explode(":", getenv('PATH')) as $directory) {
```

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
...
+    $candidate = "$dir/$cmd";
```
yorkshire-pudding:
```suggestion
    $candidate = "$directory/$command";
```

commands/db.bee.inc
```diff
@@ -348,3 +373,24 @@ function dbq_bee_callback($arguments, $options) {
     bee_message($e->getMessage(), 'error');
   }
 }
+
+/**
+ * Checks whether a command exists in the PATH, and if so, returns the full
+ * path. Returns FALSE if no such command is available.
+ *
+ * @param string $cmd
```
yorkshire-pudding:
```suggestion
 * @param string $command
```

yorkshire-pudding (changes requested):
Sorry @gwolf - a bit more to do.  After thinking about the name of the helper function, I've decided it would be better to go into `includes/filesystem.inc`.  There's also a few tweaks with the function doc bloc and variable names.

commands/db.bee.inc
gwolf: Completely makes sense. I have moved the function to the suggested file (and renamed to a more general, project-friendly way).

yorkshire-pudding:
Thanks @gwolf - this is looking good. I tested the negative condition by temporarily changing the defined executable in the `db_export_bee_callback()` function so it couldn't be found.  The function works well.

Thank you for your patience in getting this over the line.

PR #412, opened 2024-07-29
Issue #411: Add config-clear command, fix bug in config-set.

### LINKED ISSUE #411: Add 'config-clear' command and fix bug in 'config-set'
It would be nice to have the `bee config-clear file option` command to clear (remove) an option from a config file.

What's the use case? Well, if you used `bee config-set file option value` but mistyped the name of the option, you would have created an option that you can no longer remove via `bee`.

Or at least that would be possible if there were not a bug in `bee config-set`. Although the code clearly anticipates creating a nonexistent option (the `switch` statement has a branch for a NULL result for `$existing_type`), that branch is currently unreachable. So the PR for this issue fixes that bug, as well as implementing the `bee config-clear` command.

And the reason I'm doing the two things together is that we need both working functions for the test code, which I've added for both functions (creating a nonexistent option, then subsequently clearing it).

bugfolder:
Fixes https://github.com/backdrop-contrib/bee/issues/411.

### DIFF
--- a/commands/config.bee.inc
+++ b/commands/config.bee.inc
@@ -40,6 +40,20 @@ function config_bee_command() {
         'bee config-set image.style.thumbnail effects.0.data.width 200' => bt('Change the width of the Thumbnail image style.'),
       ),
     ),
+    'config-clear' => array(
+      'description' => bt('Clear the value of an option in a config file.'),
+      'callback' => 'config_clear_bee_callback',
+      'group' => 'configuration',
+      'arguments' => array(
+        'file' => bt("The name of the config object to retrieve. This is the name of the config file, less the '.json' suffix."),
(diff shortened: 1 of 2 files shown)


yorkshire-pudding (changes requested):
Thanks @bugfolder - confirmed the bug you mentioned and that this fixes it as well as the new function.

One change please:
The test results currently don't confirm that this function works:
```
Config Commands
 ✔ Config get command works
 ✔ Config set command works
 ✔ Config export command works
 ✔ Config import command works
```

Therefore, please can you split out the new test element into its own function (i.e. `public function test_config_clear_command_works()`)?

yorkshire-pudding (approved):
Thanks @bugfolder 
All looks good

PR #415, opened 2024-07-30
Issue #277: Add bee commands for theme_debug.

### LINKED ISSUE #277: Add command to enable and disable theme_debug setting
As a site architect theming my site or a theme builder
I want to be able to enable and disable the `theme_debug` setting without the `devel` module
So that I can find the template I need to override without having to use a long `config-set` command or having to have `devel` enabled.

bugfolder:
Fixes https://github.com/backdrop-contrib/bee/issues/277.


tests/phpunit.xml
```diff
@@ -23,6 +23,7 @@
       <file>backdrop/ProjectsCommandsTest.php</file>
       <file>backdrop/RolesCommandsTest.php</file>
       <file>backdrop/StateCommandsTest.php</file>
+      <file>backdrop/ThemesCommandsTest.php</file>
       <!-- There's no Status command test, since that command is used in other tests already. -->
```
yorkshire-pudding: Minor niggle - alphabetical order
```suggestion
      <!-- There's no Status command test, since that command is used in other tests already. -->
      <file>backdrop/ThemesCommandsTest.php</file>
```

yorkshire-pudding (changes requested):
Hi @bugfolder 

Thanks for working on this. I'd like to see an approach similar to the `maintenance-mode` command. Although that changes a state and this changes config, the principles are the same.  Instead of having `theme-enable-debug` and `theme-disable-debug` I would like to see `theme-debug` perhaps with an alias of `td`.  The setting would then be an argument. With `mm` I made it flexible so people could enter `0`/`1` `true`/`false` or `TRUE`/`FALSE` and it would convert to the right format for what needed setting (TRUE/FALSE in the case of `mm`, but would be 0/1 for this).

Thanks

yorkshire-pudding:
@bugfolder - did you miss this part of the review or are you still working on it?

> Thanks for working on this. I'd like to see an approach similar to the `maintenance-mode` command. Although that changes a state and this changes config, the principles are the same. Instead of having `theme-enable-debug` and `theme-disable-debug` I would like to see `theme-debug` perhaps with an alias of `td`. The setting would then be an argument. With `mm` I made it flexible so people could enter `0`/`1` `true`/`false` or `TRUE`/`FALSE` and it would convert to the right format for what needed setting (TRUE/FALSE in the case of `mm`, but would be 0/1 for this).

bugfolder:
> @bugfolder - did you miss this part of the review or are you still working on it?

No, I saw it. It basically means start over, which I'll do, but when my next time block becomes available.

Robert

PR #422, opened 2024-08-17
Add translatable string context

### LINKED ISSUE #410: Add 'context' to translatable strings to help translators
Follow on from #405 

It would make sense to identify translatable strings coming from Bee to translators as otherwise they might we left wondering which bit of the UI it relates to.

At the most basic, the `bt()` function could have a default context of simply 'Bee'.  More advanced, we could add more specific context about where in Bee the string is from.

Any thoughts?

yorkshire-pudding:
Fixes #410


yorkshire-pudding:
Oops. Branched the update branch.

PR #445, opened 2024-09-27
Add functions copied from Backdrop core module_enable() and

### LINKED ISSUE #164: UX: Indicate dependencies that are automatically enabled/disabled
When running `bee en devel_generate`, devel is also enabled because it is specified as a dependency for devel_generate 👍🏼 ...that's not being communicated to the user though:

```
bee en devel_generate

 [✔] The 'Devel Generate' module was enabled. 
```

I'd like this to be like so instead:

```
bee en devel_generate

 [i] The 'Devel' module will also be enabled, as it is required by the 'Devel Generate' module.
 [✔] The 'Devel' module was enabled.
 [✔] The 'Devel Generate' module was enabled.
```

...or:
```
bee en devel_generate

 [i] The 'Devel' module will also be enabled, as it is required by the 'Devel Generate' module.
 [i] The 'Devel' module is already enabled.
 [✔] The 'Devel Generate' module was enabled.
```

Similarly when disabling modules that are dependencies for other modules:

Current behaviour:
```
bee dis devel

 [✔] The 'Devel' module was disabled.
```

What I would like to see instead is something like this instead:
```
bee dis devel

 [i] The 'Devel Generate' module will also be disabled, as it depends on the 'Devel' module.
 [✔] The 'Devel Generate' module was disabled.
 [✔] The 'Devel' module was disabled.
```

In fact, "silently" enabling/disabling module dependencies may be seen as a bug. Drush prompts for this:
```
drush dis devel
The following projects will be disabled: devel_generate,devel.
    Do you want to disable the projects? (y/n): y

	Success: devel_generate,devel are disabled.
```

```
drush en devel_generate
The following projects will be enabled: devel, devel_generate.
    Do you want to enable the projects? (y/n): y

	Success: module devel enabled.


	Success: module devel_generate enabled.
```

elisseck:
Fixes #164

This PR works towards resolving issue #164, attempting to give better user feedback when enabling/disabling modules that have requirements/dependencies.

Currently, we are copying what Backdrop core does in the `module_enable()` and `module_disable()` functions to notify the user of dependent and/or required modules that are being disabled/enabled in addition to the named modules when calling e.g. `bee dis my_module` or `bee en my_module`.

Ideally, this is a stopgap method, and an issue will be filed against core to pop the functionality I've copied into `miscellaneous.inc`  outside of `module_enable()` and `module_disable()` so it is available to contrib modules. At that point we can use that feature and stop doing the dependency calculation twice during the process.

If there is agreement on this approach I will open the issue against Core and do the same for the install/uninstall callbacks here in `bee`. Let me know what you think!

Example output before, missing crucial information:

image

Example after:

image


yorkshire-pudding:
Hi @elisseck - If you put `Fixes #164` at the top of your description it will link to the issue (and close it, when merged)

I've had a quick look and initial feedback I would give is that messages saying 'will also be' should use the `bee_instant_message()` function rather than `bee_message()` as it's not a great look to say we will do something when it has already happened.  I will try to have a more detailed look at this very soon, but it is encouraging that the tests and PHPCS pass.

elisseck:
Cool, thank you! I'll wait for your detailed review and then I'm happy to change the `bee_message` to `bee_instant_message` for those messages with anything else; I didn't realize that was a feature :)

includes/miscellaneous.inc
```diff
@@ -206,3 +206,98 @@ function bee_format_date($timestamp, $format = 'Y-m-d H:i:s T', $timezone = NULL
...
+  // Sort the module list by pre-calculated weights.
+  asort($module_list);
+  $module_list = array_keys($module_list);
+
+  return $module_list;
+}
+
+/**
+ * Builds the full list of module requirements based on input.
+ * This function is copied from Backdrop Core module_able()
```
yorkshire-pudding:
```suggestion
 * This function is copied from Backdrop Core module_enable()
```

yorkshire-pudding (changes requested):
Hi @elisseck
Thank you for your efforts on this.

I've gone through this and the code looks mostly good (a minor typo)

I don't know if this is feasible but if I enable the submodule the messages aren't quite right. 'Entity Plus' is a dependency of 'Paragraphs'. 'Paragraphs' is the only dependency of 'Paragraphs Type Permissions' 

```
lando bee en paragraphs_bundle_permissions

 ℹ  The 'Entity Plus' module will also be enabled, as it is required by the 'Paragraphs Type Permissions' module.
 ℹ  The 'Paragraphs' module will also be enabled, as it is required by the 'Paragraphs Type Permissions' module.
 ✔  The 'Entity Plus' module was enabled.
 ✔  The 'Paragraphs' module was enabled.
 ✔  The 'Paragraphs Type Permissions' module was enabled.
```
I don't think this is a deal breaker if not as this is still a big improvement. Maybe an alternative is to clarify the message to say:
> The 'Entity Plus' module will also be enabled, as it is required by the 'Paragraphs Type Permissions' module or one of the modules it requires.

As previously mentioned, replacing `bee_message()` with `bee_instant_message()` where the message talks about what the function will do, will make more sense for the user.

Also, as it is envisaged that this will be a temporary solution until the functions are split out in core, I propose adding a new file in `includes` called `dependencies.inc` and include this in `bee.php`.  I don't know how long the core change will take so this will keep it tidy and reduce the possibilities of merge conflicts later on.

Also, it would be helpful to add a line under 'Changed' (will need adding under Unreleased) in the CHANGELOG.  Update the date for the Unreleased section to whichever date you submit the changes. I'll update if needed.

includes/miscellaneous.inc
elisseck: This got moved to `includes/dependencies.inc` but I did resolve this typo, thanks!

elisseck:
Thanks for your review @yorkshire-pudding! All makes sense to me!

I had a little trouble getting my editor to play nice with phpcs but I believe I made the changes you requested, including that typo fix although i'm not sure the formal github change request resolved since it's now in a different file.

I went with your suggestion for the note on "dependencies of dependency" module situations... the backdrop core function doesn't give us that information right now, so i'd rather not change it. We can PR the change to backdrop core moving the code as-is, and then maybe PR the change to get the dependency info we'd need to properly report this to users after that.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project follows the
 which is based on the major version of Backdrop CMS with a semantic version
 system for each contributed module, theme and layout.
 
-## [Unreleased] - 2024-09-25
+## [Unreleased] - 2024-10-09
```
yorkshire-pudding:
```suggestion
## [Unreleased] - 2024-10-10
```

yorkshire-pudding (approved):
Thanks @elisseck - looks good

